PR #2, opened 2019-05-03
single page rotation

pawelkaczmarekuk:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
+++ b/projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
@@ -6,7 +6,7 @@ import {Subscription} from 'rxjs';
 import {
   MediaViewerMessage,
   RotateDirection,
-  RotateOperation,
+  RotateOperation, RotateSinglePageOperation,
   SearchOperation,
   ZoomOperation
 } from '../../service/media-viewer-message.model';
@@ -43,12 +43,15 @@ export class PdfViewerComponent implements AfterViewInit, OnDestroy {
   }
 
   handleMessage(genericMessage: MediaViewerMessage) {
+    if (genericMessage instanceof RotateSinglePageOperation) {
+      this.rotateSinglePage(genericMessage);
+    } else
     if (genericMessage instanceof RotateOperation) {
       this.rotate(genericMessage);
-    }
+    } else
     if (genericMessage instanceof ZoomOperation) {
       this.zoom(genericMessage);
-    }
+    } else
     if (genericMessage instanceof SearchOperation) {
       this.search(genericMessage);
     }
@@ -63,6 +66,21 @@ export class PdfViewerComponent implements AfterViewInit, OnDestroy {
         currentRotation = (currentRotation + 90) % 360;
       }
       this.pdfViewer.pagesRotation = currentRotation;
+    }
+  }
+
+  rotateSinglePage(rotateSinglePageOperation: RotateSinglePageOperation) {
+    if (this.pdfViewer) {
+      const page = this.pdfViewer.getPageView(rotateSinglePageOperation.pageIndex);
+      if (page) {
+        let currentPageRotation = page.rotation;
+        if (rotateSinglePageOperation.direction === RotateDirection.LEFT) {
+          currentPageRotation = (currentPageRotation - 90) % 360;
+        } else if (rotateSinglePageOperation.direction === RotateDirection.RIGHT) {
+          currentPageRotation = (currentPageRotation + 90) % 360;
+        }
+        page.update(null, currentPageRotation);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -43,12 +43,15 @@ export class PdfViewerComponent implements AfterViewInit, OnDestroy {
   }
 
   handleMessage(genericMessage: MediaViewerMessage) {
+    if (genericMessage instanceof RotateSinglePageOperation) {
```
linusnorton: This method looks like a good candidate for https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html

My suggestion would be to have services that subscribe to particular types of message (rotate, zoom, etc) and then emit the event based on the message type. I believe the angular way is to use RX.js observables but @benouaer knows better than I do

projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -66,6 +69,21 @@ export class PdfViewerComponent implements AfterViewInit, OnDestroy {
     }
   }
 
+  rotateSinglePage(rotateSinglePageOperation: RotateSinglePageOperation) {
+    if (this.pdfViewer) {
```
linusnorton: under what circumstances is pdfViewer not set?

projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
benouaer: there is an angular library [https://ngrx.io/guide/store](ngrx) that I've mentioned to pawel that handles this kind of behaviour using 'Actions' of different types that you dispatch in different scenarios, and which are intercepted by different 'Effects' that can perform the steps you need to do.
there's some boilerplate that comes with using this library, but I think it would be worth it since there are quite a few bits of state to manage. the good thing with this is that the components are very clean, as all the state management is moved to these specialised classes.

obviously, we can always replicate the same approach ourselves if we don't have that many actions to manage

projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -66,6 +69,22 @@ export class PdfViewerComponent implements AfterViewInit, OnDestroy {
...
+      const pageView = this.pdfViewer.getPageView(rotateSinglePageOperation.pageIndex);
+      if (pageView) {
+        let currentPageRotation = pageView.rotation;
+        if (rotateSinglePageOperation.direction === RotateDirection.LEFT) {
+          currentPageRotation = (currentPageRotation - 90) % 360;
+        } else if (rotateSinglePageOperation.direction === RotateDirection.RIGHT) {
+          currentPageRotation = (currentPageRotation + 90) % 360;
+        }
```
benouaer: why not use a numeric value when invoking rotation (+90, -90), so that we don't have to check for the direction and would just add it

src/app/sandbox-webapp/sandbox-webapp.component.html
```diff
@@ -63,6 +63,9 @@
 
                   <button (click)="rotate('LEFT')">rotate left</button>
                   <button (click)="rotate('RIGHT')">rotate right</button>
+                  <button (click)="rotateSinglePage('LEFT', 0)">rotate 1st page left</button>
+                  <button (click)="rotateSinglePage('RIGHT', 0)">rotate 1st page right</button>
+
```
benouaer: why not use +90/-90 instead of Left/Right, to simplify application of rotation

projects/media-viewer/src/lib/media-viewer/viewers/pdf-viewer/pdf-viewer.component.ts
benouaer: actually, now having looked at the code a bit more closely, it seems to me that our state is contained mainly within 'pdfViewer'. so regardless of whether or not we use ngrx, if we maintained that within a store object, and then exposed functions that allow us to directly either select or update from it, that would remove the need for messages.

src/app/sandbox-webapp/sandbox-webapp.component.html
```diff
@@ -63,6 +63,9 @@
 
                   <button (click)="rotate('LEFT')">rotate left</button>
                   <button (click)="rotate('RIGHT')">rotate right</button>
+                  <button (click)="rotateSinglePage('LEFT', 0)">rotate 1st page left</button>
+                  <button (click)="rotateSinglePage('RIGHT', 0)">rotate 1st page right</button>
```
benouaer: is there a way to get the current page (based on coordinates maybe?), so that this can be used for all pages, not just the first one?

PR #6, opened 2019-05-08
Toolbar

linusnorton:
Add the toolbar as optional module

### DIFF
--- a/projects/media-viewer/src/lib/toolbar/toolbar.component.html
+++ b/projects/media-viewer/src/lib/toolbar/toolbar.component.html
@@ -1,0 +1,7 @@
+<button (click)="rotate('LEFT')">rotate left</button>
+<button (click)="rotate('RIGHT')">rotate right</button>
+<button (click)="zoom(0.2)">zoom in</button>
+<button (click)="zoom(-0.2)">zoom out</button>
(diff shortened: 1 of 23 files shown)


projects/media-viewer/src/lib/toolbar/toolbar.component.html
```diff
@@ -0,0 +1,7 @@
+<button (click)="rotate('LEFT')">rotate left</button>
+<button (click)="rotate('RIGHT')">rotate right</button>
```
benouaer: why not use +90/-90 so that we don't have check for direction later and can do the rotation direclty

src/app/routing.module.ts
```diff
@@ -4,24 +4,27 @@ import { RouterModule, Routes } from '@angular/router';
...
+    CommonModule,
+    RouterModule.forRoot(routes, {
+      scrollPositionRestoration: 'enabled',
+      anchorScrolling: 'enabled'
+    }),
+    HttpClientModule,
+    ReactiveFormsModule,
+    MediaViewerModule,
+    ToolbarModule,
+    ToolbarModule
```
benouaer: ToolbarModule seems to be imported twice

PR #11, opened 2019-05-09
toolbar moved inside media-viewer

pawelkaczmarekuk:
### Change description ###

Toolbar moved and fed with URL and media type


projects/media-viewer/src/lib/media-viewer/media-viewer.component.ts
```diff
@@ -20,9 +20,11 @@ export class MediaViewerComponent implements OnInit {
   private supportedContentTypes = ['pdf', 'image'];
 
   ngOnInit(): void {
-    this.rotation = this.actionEvents.rotate;
-    this.search = this.actionEvents.search;
-    this.zoom = this.actionEvents.zoom;
+    if (this.actionEvents) {
```
linusnorton: when is actionEvents not specified?

pawelkaczmarekuk: for unsupported type viewer

PR #12, opened 2019-05-09
Download

pawelkaczmarekuk:
### Change description ###

download functionality


projects/media-viewer/src/lib/media-viewer/toolbar/toolbar.component.ts
```diff
@@ -10,11 +11,12 @@ export class ToolbarComponent {
 
   @Input() url = '';
   @Input() contentType: string;
+  @Input() downloadFileName = null;
   sidebarToggle = false;
   searchToggle = false;
   secondaryToolbarToggle = false;
 
-  constructor() {}
+  constructor(private pdfWrapper: PdfJsWrapper) {}
```
linusnorton: I don't think the toolbar should depend on the wrapper directly, it should probably fire off an event somewhere in the same manner as the action events - what do you think?

benouaer:
download seems to already be done with the other PR https://github.com/hmcts/media-viewer/pull/13/files

PR #14, opened 2019-05-09
split toolbar into smaller more manageable components

benouaer:
### Change description ###
splitting the toolbar into smaller more manageable components
```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/toolbar/findbar.component.ts
```diff
@@ -0,0 +1,56 @@
+import { Component, Input } from '@angular/core';
+import { ActionEvents, SearchOperation } from '../media-viewer/media-viewer.model';
+
+@Component({
+  selector: 'app-findbar',
+  template: `
```
linusnorton: This seems like a lot for an inline template - how come you didn't use a file?

benouaer: yes, tbh, what I really wanted was to have templates that you can just include. but if it looks untidy, I can create separate html files

linusnorton: Personally I would find it easier to modify the template if it was in it's own file

benouaer: I agree tbh, I'll change it

PR #15, opened 2019-05-13
cleanup

pawelkaczmarekuk:
### Change description ###

slightly more consistent take on messaging

### DIFF
--- a/projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
+++ b/projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
@@ -22,26 +22,30 @@ export class ImageViewerComponent {
 
   @Input()
   set rotateOperation(operation: RotateOperation) {
+    if (operation) {
       this.rotation = this.rotation + operation.rotation;
       this.rotateImage();
+    }
   }
 
   @Input()
   set zoomOperation(operation: ZoomOperation) {
-
+    if (operation) {
+      // TODO
+    }
   }
 
   @Input()
-  set printOperation(printOperation: PrintOperation) {
-    if (printOperation) {
+  set printOperation(operation: PrintOperation) {
+    if (operation) {
       const printWindow = window.open(this.url);
       printWindow.print();
     }
   }
 
   @Input()
-  set downloadOperation(downloadOperation: DownloadOperation) {
-    if (downloadOperation) {
+  set downloadOperation(operation: DownloadOperation) {
+    if (operation) {
       // TODO
     }
   }
--- a/projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
+++ b/projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
@@ -21,7 +21,7 @@ export class ImageViewerComponent {
 
 
   @Input()
-  set rotateOperation(operation: RotateOperation) {
+  set rotateOperation(operation: RotateOperation | null) {
     if (operation) {
       this.rotation = this.rotation + operation.rotation;
       this.rotateImage();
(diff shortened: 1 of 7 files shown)


projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
```diff
@@ -1,52 +1,60 @@
...
-      if(operation && operation.action) {
-        this[operation.action].call(this, operation);
-      }
-    }
-  }
 
-  rotate(operation: RotateOperation) {
+  @Input()
+  set rotateOperation(operation: RotateOperation) {
+    if (operation) {
```
linusnorton: this is a bit nonsensical - if the type declaration of operation is RotateOperation then it will always exist. This was not true in typescript 2.x (and you can disable it in 3.x) but types are non-nullable by default. If you want a nullable type the declaration needs to be `RotateOperation | null`

linusnorton: if you want an optional type then you add `?` which is the same as `| undefined`

linusnorton:
On the whole it's much better but these `if` statements everywhere for `operation` are unnecessary

projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
benouaer: the first time these functions are called, when the component is first loaded, the value seems to be null

pawelkaczmarekuk: set rotateOperation is a way to define a setter - I can't make the parameter optional. I added the if statement because it's executed with the argument == null - I'm guessing when it's initialised.

pawelkaczmarekuk: so without the if statement the print would execute when the component was being initialised. Any ideas how to make it prettier ?

linusnorton: personally, I would look into why a null is being emitted... that is super weird

linusnorton: if you absolutely can't work it out, then change the type definition to reflect that the value might be null

benouaer: it's not that a null is being emitted. it's the async pipe evaluating to null when the subject value hasn't resolved yet

linusnorton: Why would it emit something before it's being resolved?

benouaer: the subject isn't emitting null. it's the angular async pipe that returns null when no value has been received from the subject yet.
here's a proposal/bug that's been raised to try and get them to change this behaviour
https://github.com/angular/angular/issues/16982

linusnorton: leaky abstraction much? okay, let's update the type info so it makes sense then merge it

pawelkaczmarekuk: would it be cleaner to have non-null pipe as a workaround? 

[zoomOperation]="zoom | async | non-null"

linusnorton: if that's possible it would be good - I've  just pushed a change to update the type info

PR #20, opened 2019-05-14
Search improvements

linusnorton:
Search functionality

### DIFF
--- a/projects/media-viewer/src/lib/media-viewer/media-viewer.model.ts
+++ b/projects/media-viewer/src/lib/media-viewer/media-viewer.model.ts
@@ -1,8 +1,10 @@
 import { Subject } from 'rxjs';
+import { debounceTime } from 'rxjs/operators';
 
 export class ActionEvents {
   public readonly rotate = new Subject<RotateOperation>();
-  public readonly search = new Subject<SearchOperation>();
+  public readonly search = new Subject<SearchOperation>().pipe(debounceTime(250));
+  public readonly searchResultsCount = new Subject<SearchResultsCount>();
   public readonly zoom = new Subject<ZoomOperation>();
   public readonly print = new Subject<PrintOperation>();
   public readonly download = new Subject<DownloadOperation>();
@@ -32,3 +34,7 @@ export class SearchOperation {
   ) {}
 }
 
+export interface SearchResultsCount {
+  current: number;
+  total: number;
+}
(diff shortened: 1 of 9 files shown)


projects/media-viewer/src/lib/media-viewer/media-viewer.model.ts
```diff
@@ -1,8 +1,10 @@
 import { Subject } from 'rxjs';
+import { debounceTime } from 'rxjs/operators';
 
 export class ActionEvents {
   public readonly rotate = new Subject<RotateOperation>();
-  public readonly search = new Subject<SearchOperation>();
+  public readonly search = new Subject<SearchOperation>().pipe(debounceTime(250));
```
benouaer: why do we need the debounceTime?

linusnorton: the event is triggered on every keypress and the PDF lib struggles to keep up when you type quickly

projects/media-viewer/src/lib/media-viewer/toolbar/search-bar/search-bar.component.ts
```diff
@@ -10,21 +10,59 @@ import { BehaviorSubject } from 'rxjs';
...
+  resultsText = '';
+  searchText = '';
+  haveResults = false;
 
   constructor() {}
 
-  searchNext(searchTerm: string) {
-    this.actionEvents.search.next(new SearchOperation(searchTerm, false, searchTerm !== this.lastSearchTerm));
+  ngOnInit() {
+    this.searchResultsCount.subscribe(r => this.onUpdateResults(r));
```
benouaer: it might be easier to use the async pipe to avoid having to subscribe to this. then we can just have a setter method that reacts to new values coming in, in the same way onUpdateResults is doing that

linusnorton: I tried to do that but failed, for some reason when I used set it sent in the observable rather than the operation, I was muchos confused.

projects/media-viewer/src/lib/media-viewer/toolbar/toolbar.component.html
```diff
@@ -1,7 +1,8 @@
 <div id="outerContainer" [ngClass]="{ sidebarOpen: sidebarOpen | async }">
   <mv-side-bar id="sidebarContainer"></mv-side-bar>
   <div id="mainContainer">
-    <mv-search-bar [actionEvents]="actionEvents"
+    <mv-search-bar [searchEvents]="actionEvents.search"
+                   [searchResultsCount]="actionEvents.searchResultsCount"
```
benouaer: we could use the async pipe here, as mentioned above

PR #21, opened 2019-05-14
download and print

pawelkaczmarekuk:
### Change description ###

Implementation of print and download


projects/media-viewer/src/lib/media-viewer/viewers/image-viewer/image-viewer.component.ts
```diff
@@ -38,15 +40,20 @@ export class ImageViewerComponent {
...
   set downloadOperation(operation: DownloadOperation | null) {
     if (operation) {
-      // TODO
+      const a = document.createElement('a');
+      document.body.appendChild(a);
+      a.setAttribute('style', 'display: none');
+      a.href = this.url;
+      a.download = this.downloadFileName;
+      a.click();
+      a.remove();
```
linusnorton: I'm assuming there is no better way to do this?

pawelkaczmarekuk: it seems like the most reliable solution...

PR #31, opened 2019-05-20
Switch to the double size icons

linusnorton:


projects/media-viewer/src/lib/media-viewer/styles/media.scss
```diff
@@ -30,10 +30,10 @@
     background-size: 16px 17px;
   }
 
-  .dropdownToolbarButton {
-    background: url(/assets/images/toolbarButton-menuArrows@2x.png) no-repeat;
-    background-size: 7px 16px;
-  }
+  //.dropdownToolbarButton {
```
benouaer: this is no longer needed?

PR #39, opened 2019-05-22
changing defaults

pawelkaczmarekuk:
### Change description ###

I moved the init of ActionEvents to media-viewer. It should be an optional action if a consumer chooses not to use the toolbar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ component entry point:
 </mv-media-viewer>  
 ```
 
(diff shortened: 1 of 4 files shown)


linusnorton:
So it can still be set if required?

pawelkaczmarekuk:
> So it can still be set if required?

yes.

linusnorton:
Could you update the README.md to state that it is now optional

pawelkaczmarekuk:
> Could you update the README.md to state that it is now optional

Done.

PR #40, opened 2019-05-22
attempt to fix a worker

pawelkaczmarekuk:
### Change description ###

An attempt to fix the Worker. Let me know what you think. I added RequireJs to the module. Not sure if it's not better to leave to the client to decide whether to use a RequireJs or not to use it and then accept using a Fake worker with worse performance...


package.json
```diff
@@ -18,7 +18,8 @@
     "copy-hmcts-assets": "rsync -r projects/media-viewer/node_modules/@hmcts/frontend/* dist/media-viewer/assets/hmcts-frontend",
     "copy-govuk-assets": "rsync -r projects/media-viewer/node_modules/govuk-frontend dist/media-viewer/assets",
     "install-js-dependencies": "cd projects/media-viewer && npm install",
-    "copy-files": "npm run copy-assets && npm run copy-docs",
+    "copy-js-dependencies": "rsync -r projects/media-viewer/node_modules/pdfjs-dist/build dist/media-viewer/assets",
+    "copy-files": "npm run copy-js-dependencies && npm run copy-assets && npm run copy-docs",
```
linusnorton: do we still need this now you've taken it out of the assets?

PR #42, opened 2019-05-23
Test/e2e

velimd:
included functional test for media viewer


e2e/protractor.conf.js
```diff
@@ -2,14 +2,23 @@
 // https://github.com/angular/protractor/blob/master/lib/config.ts
 
 const { SpecReporter } = require('jasmine-spec-reporter');
+var fs = require('fs');
```
linusnorton: any reason you went for a var over const?

e2e/src/scenarios/download.e2e-spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+  it('should download the pdf', () => {
+    const file = 'e2e/src/downloads/example.pdf';
+
+    page.getPdfViewer().click();
+
+    const downloadPage = page.getDownloadButton();
+    downloadPage.click()
+      .then(() => page.wait(3000))
```
linusnorton: should we use await rather than .then?

velimd: youre right, then again, i dont think the then should be there as page.wait would do the same thing

e2e/protractor.conf.js
velimd: nope, just was looking into how to read files in typescript. will change it.

e2e/src/scenarios/download.e2e-spec.ts
linusnorton: Well the .then comment applies to the whole PR really. I don't think we should be using them when we have `await`. 

The page.wait function will probably still be needed as that's to wait for a specific amount of time (I assume to let the HTML update).

velimd: yh for html update and in this case its to wait for the file to be downloaded

e2e/protractor.conf.js
```diff
@@ -24,5 +33,23 @@ exports.config = {
...
     });
     jasmine.getEnv().addReporter(new SpecReporter({ spec: { displayStacktrace: true } }));
+    removePreviousDownloads();
   }
 };
+
+function removePreviousDownloads() {
+  try {
+    const downloadsPath = 'e2e/src/downloads';
+    const files = fs.readdirSync(downloadsPath);
```
JeroenRijks: Is this an array of relative file paths, or an array of actual File objects? If it's the first, could you rename to something like relativeFilePaths?

e2e/src/app.po.ts
```diff
@@ -1,27 +1,124 @@
...
+
+  getCurrentZoomOption() {
+    return element(by.id('customScaleOption'));
+  }
+
+  getZoomSelect() {
+    return element(by.id('scaleSelect'));
+  }
+
+  selectZoomValue(zoomOption) {
```
JeroenRijks: this, and other methods here, are still getters right? getSelectZoomValue?

e2e/src/scenarios/print.e2e-spec.ts
```diff
@@ -0,0 +1,44 @@
...
+    const printPage = page.getPrintButton();
+    printPage.click();
+
+    // clean up by moving to app.po
+    browser.getAllWindowHandles().then(function(handles) {
+      expect(handles.length).toEqual(2);
+      browser.switchTo().window(handles[1]);
+      expect(page.getPrintDialog).toBeTruthy();
+      page.wait(1000);
+      browser.close();
```
JeroenRijks: weird how you close the browser, then switch to another tab in the browser. Is there any way to do something like `browser.close(window(handles[1]))`?

JeroenRijks: nvm, I couldn't find anything better than what you've got

e2e/protractor.conf.js
velimd: it returns an array with the files names rather than paths.

PR #53, opened 2019-05-30
DO NOT MERGE attempt of design of anno components

pawelkaczmarekuk:
### Change description ###

Attempt of design of annotation components - it's just scratches the surface need more work.


projects/media-viewer/src/lib/media-viewer/annotation/annotation-set.model.ts
```diff
@@ -0,0 +1,127 @@
...
+  createdBy: string;
+  createdByDetails: IdamDetails;
+  createdDate: any;
+  lastModifiedBy: string;
+  lastModifiedByDetails: IdamDetails;
+  lastModifiedDate: any;
+  content: string;
+}
+
+export class Comment implements IComment {
```
benouaer: should we put the classes/interfaces in separate locations, so that we can export the interfaces separately from the concrete implementations

projects/media-viewer/src/lib/media-viewer/annotation/annotation-set.model.ts
```diff
@@ -0,0 +1,127 @@
...
+    public page?: number,
+    public color?: string,
+    public comments?: Comment[],
+    public rectangles?: Rectangle[],
+    public type?: string) {
+  }
+}
+
+export class IdamDetails {
+  constructor(
```
benouaer: is this needed?

projects/media-viewer/src/lib/media-viewer/annotation/annotation-set.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  save(annotationSet: AnnotationSet): void;
+
+  delete(annotation: Annotation): void;
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AnnotationSetService implements IAnnotationSetService {
```
benouaer: again, it seems like the class and interface should be separated so that we export the interface and not the class, or at least that would be done independently

projects/media-viewer/src/lib/media-viewer/annotation/annotation-set.model.ts
```diff
@@ -0,0 +1,127 @@
...
+    public id?: string,
+    public annotationSetId?: string,
+    public createdBy?: string,
+    public createdDate?: any,
+    public createdByDetails?: IdamDetails,
+    public lastModifiedBy?: string,
+    public lastModifiedByDetails?: IdamDetails,
+    public lastModifiedDate?: any,
+    public documentId?: string,
+    public page?: number,
```
benouaer: not sure, but this feels like it should be separated into a UI bit, which has page, color, rectangle, and the other bits to do with the user would be on a wrapper object. just a thought

PR #56, opened 2019-06-04
Annotation structure

linusnorton:
Basic annotation data structures

### DIFF
--- a/projects/media-viewer/src/lib/media-viewer/annotations/annotation-api.service.ts
+++ b/projects/media-viewer/src/lib/media-viewer/annotations/annotation-api.service.ts
@@ -1,0 +1,37 @@
+import { Observable } from 'rxjs';
+import { HttpClient, HttpResponse } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { AnnotationSet } from './annotation-set.model';
+import { Annotation } from './annotation.model';
+
+@Injectable()
+export class AnnotationApiService {
+
+  constructor(
+    private readonly httpClient: HttpClient
+  ) { }
+
+  public createAnnotationSet(body: Partial<AnnotationSet>): Observable<HttpResponse<AnnotationSet>> {
+    const url = '/em-anno/annotation-sets';
+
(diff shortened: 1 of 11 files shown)


projects/media-viewer/src/lib/media-viewer/annotations/annotation-api.service.ts
```diff
@@ -0,0 +1,37 @@
...
+    const url = `/em-anno/annotations/${annotation.id}`;
+
+    return this.httpClient.delete<Annotation>(url, { observe: 'response' });
+  }
+
+  public createAnnotation(annotation: Annotation): Observable<HttpResponse<Annotation>> {
+    const url = `/em-anno/annotations`;
+
+    return this.httpClient.post<Annotation>(url, annotation, { observe: 'response' });
+  }
```
velimd: All fine, just a questions, would we also need edit as well as if a comment changed or the annotations position when its dragged/resized

projects/media-viewer/src/lib/media-viewer/annotations/annotations.module.ts
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  entryComponents: [
+  ],
+  providers: [
+    AnnotationApiService
+  ],
+  exports: [
+  ]
+})
+export class AnnotationsModule { }
```
velimd: i did this however, as my changes arent here anymore, maybe you want to add this module in the media-viewer module?

projects/media-viewer/src/lib/media-viewer/annotations/annotation-api.service.ts
linusnorton: I think we will. I did not see it in the document-viewer api code though

projects/media-viewer/src/lib/media-viewer/annotations/annotations.module.ts
linusnorton: ok

PR #60, opened 2019-06-07
move the assets back

linusnorton:
Angular requires them to be under src/


velimd:
kinda sucks angular requires that, might be why we had media viewer inside before.

linusnorton:
it won't let you require something outside of the src/ folder

PR #61, opened 2019-06-07
refreshing the viewer when change has occurred.

velimd:
including ngOnChange when the input value has changed


linusnorton:
what purpose does this serve?

velimd:
> what purpose does this serve?

to update the viewer, when the input change but seems like i dont need it now.

PR #63, opened 2019-06-11
Import annotations

benouaer:
https://tools.hmcts.net/jira/browse/EM-1342
### Change description ###
show annotation in the media viewer
```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/annotations/comment/comment.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  set comment(comment: Comment) {
+    this.lastUpdate = comment.lastModifiedDate ? comment.lastModifiedDate : comment.createdDate;
+    this.author = comment.createdByDetails;
+    this.editor = comment.lastModifiedByDetails;
+    this.content = comment.content;
+  }
+
+
+  ngOnInit() {
+    this.editable = false;
```
linusnorton: why are these set here rather than where they are defined?

projects/media-viewer/src/lib/annotations/annotations.component.html
```diff
@@ -0,0 +1,15 @@
...
+  <div class="annotation-wrapper" *ngFor="let annotation of annotations">
+    <div [ngStyle]="{ top: annotation.rectangles[0].y + 'px' }">
+      <mv-anno-rectangle
+        [rectangle]="annotation.rectangles[0]"
+        [color]="annotation.color">
+      </mv-anno-rectangle>
+      <mv-anno-comment
+        [comment]="annotation.comments[0]"
+        [user]="annotation.createdByDetails"
+        [height]="annotation.rectangles[0].y">
```
linusnorton: why is height set to the y value rather than the height?

PR #65, opened 2019-06-12
reset state between e2e tests

JeroenRijks:
Not applicable

### Change description ###
reset state between every e2e test, to ensure that they do not depend on each other. Also, since they don't depend on each other, I've set them up to all run simultaneously.

```
[ ] Yes
[x] No
```


projects/media-viewer/src/lib/comments-summary/comments-summary.component.html
```diff
@@ -1,7 +1,7 @@
-<div class="modal" (click)="onClose()">
-  <div class="modal-content govuk-width-container clearfix" (click)="$event.stopPropagation()">
+<div id="modal-background" class="modal" (click)="onClose()">
```
JeroenRijks: The changes to this file will be used in another PR - it will allow the e2e tests to "click" on stuff, for the comment-summary feature

PR #70, opened 2019-06-14
Delete comment

linusnorton:


projects/media-viewer/src/lib/annotations/annotations.component.spec.ts
```diff
@@ -30,6 +30,10 @@ describe('AnnotationsComponent', () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(AnnotationsComponent);
     component = fixture.componentInstance;
+
+    component.annotationSet = JSON.parse(JSON.stringify(annotationSet));
```
JeroenRijks: what's this for?

linusnorton: Cloning the object as the delete test mutates it

PR #72, opened 2019-06-17
Edit comments

linusnorton:
Edit comments

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotations.component.spec.ts
+++ b/projects/media-viewer/src/lib/annotations/annotations.component.spec.ts
@@ -1,19 +1,25 @@
 import { AnnotationsComponent } from './annotations.component';
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { CommentComponent } from './comment/comment.component';
 import { RectangleComponent } from './rectangle/rectangle.component';
 import { FormsModule } from '@angular/forms';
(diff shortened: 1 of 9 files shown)


projects/media-viewer/src/lib/annotations/annotations.component.spec.ts
```diff
@@ -70,20 +70,30 @@ describe('AnnotationsComponent', () => {
...
+      const buttons = commentElements[1].querySelectorAll('button');
       buttons[1].click();
     });
-
   });
 
+  it('updates a comment', async () => {
+    await new Promise(resolve => {
+      component.update.subscribe(as => {
+        expect(as.annotations[0].comments[0].content).toBe('Updated comment');
```
JeroenRijks: 'as' confused me for a sec, maybe call it annotationSet?

JeroenRijks: I guess from context, it's not hard to figure out.

projects/media-viewer/src/lib/annotations/annotations.component.spec.ts
```diff
@@ -70,20 +70,30 @@ describe('AnnotationsComponent', () => {
...
+
+        resolve();
+      });
+
+      component.updateComment(0, 'Updated comment');
```
JeroenRijks: is the previous section setting up a promise, which doesn't get resolved until this runs? What happens if this line isn't run, does the test fail, or just stay open indefinitely? Is there a timeout or something?

PR #79, opened 2019-06-20
fix issues with annotation selection

benouaer:
### Change description ###
fix issues with annotation selection
```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotation.component.html
+++ b/projects/media-viewer/src/lib/annotations/annotation.component.html
@@ -1,7 +1,7 @@
-<div class="annotation-wrapper">
+<div class="annotation-wrapper" tabindex="1" (focusout)="toggleSelection(false)">
   <mv-anno-rectangle
-    (click)="selectAnnotation()"
-    [selected]="(selectedAnnotation | async) === annotation.id"
+    (click)="toggleSelection(true)"
+    [selected]="selected"
     [zoom]="zoom"
     [rectangle]="annotation.rectangles[0]"
     [draggable]="draggable"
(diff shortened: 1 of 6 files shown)


projects/media-viewer/src/lib/annotations/annotation.component.html
```diff
@@ -1,18 +1,18 @@
-<div class="annotation-wrapper">
+<div class="annotation-wrapper" tabindex="1" (focusout)="toggleSelection(false)">
```
linusnorton: focusout only works in input/textarea fields - if you select the title of the comment then try to click somewhere else it probably won't unselect

projects/media-viewer/src/lib/viewers/pdf-viewer/annotations-view.injector.ts
```diff
@@ -14,7 +14,6 @@ export class AnnotationsViewInjector {
       this.viewContainerRef.clear();
       const pdfViewerHtml = pdfViewer.nativeElement as HTMLElement;
       let pages = pdfViewerHtml.querySelectorAll(".page");
-      const selectedAnnotationEvent = new Subject<string>();
 
       if (pages.length && pages.length > 0) {
         annotations.forEach(annotation => {
```
linusnorton: Thought I asked to get rid of this .forEach

benouaer: is there a reason to use a for-loop instead of a for-each

linusnorton: I guess it's just a matter of style, for me `for (const annotation of annotations)` is more natural as it uses language constructs rather than higher order functions. It performs better as well, but I guess that doesn't matter

benouaer: ok, happy to change it

benouaer: done

PR #88, opened 2019-06-26
Completed the HighlightMode Event Generation, refactored and provided…

russellwardink:
Completed the HighlightMode Event Generation, refactored and provided a toolbar and viewer events service that can be injected into components instead of passing Subjects around the component tree. Migrated the events folder into a shared directory (DDD) Concepts along with ensuring all necessary references were updated.

Committing now as a neat refactor step, before continuing on a new branch with the feature.

EM 1711 (Part of)

```
[  ] Yes
[ X ] No (However, take note of newly deprecated methods and properties.
```


projects/media-viewer/src/lib/shared/toolbar.events.service.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  beforeEach(() => {
+    service = new ToolbarEventsService();
+    service.drawMode.next(false);
+    service.highlightMode.next(false);
+  });
+  afterEach(() => {
+    service = null;
+  });
+  it('should create', () => {
+        expect(service).toBeTruthy();
```
linusnorton: what's up with the formatting here?

projects/media-viewer/src/lib/shared/toolbar-button-toggles.ts
```diff
@@ -16,7 +16,6 @@ export class ToolbarButtonToggles {
   public readonly sidebarOpen = new BehaviorSubject(false);
   public readonly searchBarHidden = new BehaviorSubject(true);
   public readonly drawMode = new BehaviorSubject(false);
```
linusnorton: I'm a bit confused, shouldn't draw mode die as well?

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
```diff
@@ -0,0 +1,36 @@
...
+export class ToolbarEventsService {
+  // Register Observable Subject Events relevant to the Toolbar
+  highlightMode = new BehaviorSubject<HighlightMode>(false);
+  drawMode = new BehaviorSubject<DrawMode>(false);
+  constructor() {}
+
+  // Function to inform Observers that highlightMode has been enabled
+  public toggleHighlightMode(): void {
+    // Highlight and Draw states are mutually exclusive
+    if (this.highlightMode.getValue() === false) {
```
benouaer: maybe this could be simplified by emitting the negative of the current value
this.drawMode.next(!this.highlightMode.getValue());
this.highlightMode.next(this.highlightMode.getValue());

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
```diff
@@ -0,0 +1,36 @@
...
+// Toolbar Custom-Event Types
+export type HighlightMode = boolean;
+export type DrawMode = boolean;
+
+@Injectable({providedIn: 'root'})
+export class ToolbarEventsService {
+  // Register Observable Subject Events relevant to the Toolbar
+  highlightMode = new BehaviorSubject<HighlightMode>(false);
+  drawMode = new BehaviorSubject<DrawMode>(false);
+  constructor() {}
```
linusnorton: can be removed

projects/media-viewer/src/lib/shared/viewer-operations.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * @deprecated Please DO NOT ADD to this file
+ * Instead please add new Operations to viewer.events.service.ts
+ */
```
benouaer: maybe we could keep these operations defined here and use them within the service

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
```diff
@@ -0,0 +1,36 @@
+import {Injectable} from '@angular/core';
+import {BehaviorSubject} from 'rxjs';
+
+// Toolbar Custom-Event Types
+export type HighlightMode = boolean;
+export type DrawMode = boolean;
+
+@Injectable({providedIn: 'root'})
+export class ToolbarEventsService {
```
linusnorton: I'm not 100% sure on the angular naming convention but I think this file should be called `toolbar-events.service.ts` - it looks a bit weird but I think it's convention.

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: the issue with the above is the mutually exclusive behaviour. Its not a rocker switch, but instead there are multiple states. On / Off, and if you click on the other ones, they turn the respective other toggle on or off. Using negation on the current value doesn’t help to understand that logic.

projects/media-viewer/src/lib/shared/viewer.events.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  x: number;
+  y: number;
+  height: number;
+  width: number;
+}
+export type TextSelected = TextSelectionRectangle[];
+
+
+@Injectable({providedIn: 'root'})
+export class ViewerEventsService {
```
linusnorton: Same as above re filename. I'm not really fussed though

projects/media-viewer/src/lib/shared/toolbar-button-toggles.ts
russellwardink: Yes, however i'm not confident that it's all it does, and i'd be refactoring most of the app.
Its on my target list to get rid of, but in order to get the skeleton in place as requested by everyone this morning, its a 'leave it alone' until i can be sure it doesn't break anything else.

projects/media-viewer/src/lib/shared/viewer.events.service.ts
```diff
@@ -0,0 +1,25 @@
+import {Injectable} from '@angular/core';
+import {Subject} from 'rxjs';
+
+
+// Viewer Custom-Event Types
+export interface TextSelectionRectangle {
+  x: number;
```
benouaer: this could get very large very quickly if we include the whole domain in here. might be worth splitting into separate files

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: why?

russellwardink: or the constructor you mean?

projects/media-viewer/src/lib/shared/viewer-operations.ts
russellwardink: Potentially, but they remain where they are for now. That is certainly up for discussion, hence the deprecation.

projects/media-viewer/src/lib/shared/viewer.events.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+
+@Injectable({providedIn: 'root'})
+export class ViewerEventsService {
+  // Register Observable Subject Events relevant to the Viewers
+  textSelected = new Subject<TextSelected>();
+  constructor() {}
+
+  // Function to inform Observers that text has been selected in the viewer
+  public onTextSelection(selectionData: TextSelected): void {
```
linusnorton: I'm  not sure about this file...  the service approach makes sense for `ToolbarEventsService` as it's encapsulating the state like a repository would. This feels a bit different though because it's encapsulating what should be an `@Output()` event.

projects/media-viewer/src/lib/shared/toolbar.events.service.spec.ts
russellwardink: just pushed a fix for this

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
```diff
@@ -18,10 +19,41 @@ export class ToolbarLeftPaneComponent {
...
+  @Input() highlightMode = false;
+  @Input() drawMode = false;
+
+  // local array of any subscriptions so that we can tidy them up later
+  private subscriptions: Subscription[] = [];
+  constructor(private readonly toolbarEventsService: ToolbarEventsService) {}
+
+  ngOnInit(): void {
+    // Listen for any changes invoked on the toolbar events Service and initialise any default behaviour state
+    this.subscriptions.push(this.toolbarEventsService.highlightMode.subscribe((toggleValue) => {
```
linusnorton: I would re-iterate my objection from our call earlier.... it seems we're adding a lot of boiler plate (that is will duplicated in other classes too) in  order to avoid `| async`.

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: I agree, however @linusnorton suggested to place them where used. I’m happy to put these in a Types file again later on (in the same folder) if they grow too big

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
```diff
@@ -18,10 +19,41 @@ export class ToolbarLeftPaneComponent {
...
+    }));
+    this.subscriptions.push(this.toolbarEventsService.drawMode.subscribe((toggleValue) => {
+      this.drawMode = toggleValue;
+    }));
+  }
 
-  constructor() {}
+  ngOnDestroy(): void {
+    // Clean up any subscriptions that we may have
+    this.subscriptions.forEach((subscription) => {
```
linusnorton: I mentioned this to Taleb the other day - I prefer to use the `for (const subscription of this.subscriptions) { ... }` approach. There is not much between them but I prefer to use a language construct rather than a higher order function. It feels more natural.

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
```diff
@@ -10,14 +10,15 @@
...
                        [showSidebarToggleBtn]="toolbarButtons?.showSidebarToggleBtn"
                        [showSearchbarToggleBtn]="toolbarButtons?.showSearchbarToggleBtn"
                        [showNavigationBtns]="toolbarButtons?.showNavigationBtns"
                        [searchBarHidden]="toolbarButtons?.searchBarHidden"
                        [showHighlightBtn]="toolbarButtons?.showHighlightBtn"
-                       [drawMode]="toolbarButtons?.drawMode"
-                       [sidebarOpen]="toolbarButtons?.sidebarOpen"></mv-tb-left-pane>
+                       [sidebarOpen]="toolbarButtons?.sidebarOpen"
+                       [highlightMode]="false"
+                       [drawMode]="false"
```
linusnorton: how is this  going to work?

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: I see why you are suggesting that, however, I see a need to act on Viewer Events from other components… e.g the Toolbar, or alternatively if we split the comments out, that will too need to become viewer event aware…

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.ts
```diff
@@ -37,12 +38,28 @@ export class ImageViewerComponent implements OnChanges {
...
   constructor(
     private readonly printService: PrintService,
-    private readonly api: AnnotationApiService
+    private readonly api: AnnotationApiService,
+    private readonly toolbarEventsService: ToolbarEventsService
   ) { }
 
+  ngOnInit(): void {
+    // Listen for any changes invoked on the toolbar events Service and initialise any default behaviour state
+    this.subscriptions.push(this.toolbarEventsService.drawMode.subscribe((toggleValue) => {
```
linusnorton: oh look... it's that boilerplate again :D it's feeling a bit wet in here.

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
benouaer: yes, it's less obvious, but the second line gives you a hint that the values are mutually exclusive

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
linusnorton: yeah, if it doesn't do anything you can just remove  it

projects/media-viewer/src/lib/shared/viewer.events.service.ts
linusnorton: I would suggest `annotations/annotation-set/text-selection-rectangle.model.ts`

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
russellwardink: I hear you, but on the contrary, i disagree in that. I believe it makes it far clearer that we are listening for key events, and whilst its more explicit, it does make it easier to debug and walk through the logic.

In this instance, its changing an internal property that is only being used by the view template. So this is a good candidate to use the async pipe…

projects/media-viewer/src/lib/shared/viewer.events.service.ts
linusnorton: okay, I'd like to know what @benouaer thinks

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: i agree, it probably should be a -. I'll revise.

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: agreed, but will revise

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
russellwardink: Not sure yet, it was one of the Seams that I highlighted could be shut off without causing harm. It does need tidying up though on the next step.

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.ts
russellwardink: lol...

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
benouaer: do we need to pass these through given that they're now coming from the toolbarEventsService

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
benouaer: could we use the async pipe here, as it's only for view state

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: ok cool, will remove that

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: ok cool

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: not really no..
So scenario 1: 
They act as a toggle. Your suggested code works in that instance.
Scenario 2: 
Your code falls over as the draw mode will always be the opposite of the HighlightButton state, it doesn’t account for the instances where you want both off.

I can’t think of another way of achieving this bar a switch statement, so going to keep as is, as this breaks my tests otherwise

projects/media-viewer/src/lib/shared/toolbar-button-toggles.ts
russellwardink: gone through and killed this to keep things clean, will be corrected in a coming commit

projects/media-viewer/src/lib/shared/toolbar.events.service.spec.ts
russellwardink: @linusnorton can you recheck?

projects/media-viewer/src/lib/shared/toolbar-button-toggles.ts
russellwardink: have cleaned this up now @linusnorton can you re-check?

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: constructor removed. @linusnorton can you recheck?

projects/media-viewer/src/lib/shared/toolbar.events.service.ts
russellwardink: renamed. @linusnorton can you recheck?

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: renamed. @linusnorton can you recheck?

projects/media-viewer/src/lib/shared/viewer.events.service.ts
russellwardink: until i implement this a bit further post this merge, going to keep it in situ. As not settled on how to tap into the Rectangle implementation yet. For the moment, its purely a type to determine a TextSelection event

russellwardink: @taleb @linusnorton are you both ok with this for now?

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
russellwardink: @linusnorton changed to your suggestion. Can you re-check.
Overall, i’m not too fussed about developer styles within the team on choice of control flow. Ultimately there are numerous ways of iterating and it comes down to preference. Maybe we need to look at a code linter config so that we pick these up as we right if its deemed important.

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
russellwardink: @benouaer removed, can you recheck?

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
linusnorton: Was just adding my preference but it doesn't matter ultimately.

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
russellwardink: i’ve conceded in this example… its now an async pipe. @linusnorton can you recheck?

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
russellwardink: @linusnorton are you ok with this for now?

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
russellwardink: i’ve conceded in this example… its now an async pipe. @benouaer can you recheck?

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.ts
russellwardink: in this scenario… i believe this is the right pattern. Its responsible for updating an internal property within the component. 
I can’t see this being an async pipe scenario. @linusnorton this is the pattern i'd like for us to use in scenarios where state of the component is being changed

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
russellwardink: actually removed them for the time being

PR #89, opened 2019-06-27
Fix selection

linusnorton:


linusnorton:
Russell has given this a thumbs up in another PR

PR #90, opened 2019-06-27
fix comment editing

linusnorton:
This PR changes the mouse event handling in the AnnotationSet component so that the comment editing correctly focuses. I'm not quite sure why an event handler higher in the event change was able to block one lower down but that's the way it seemed to work.


linusnorton:
@velimd I've not found a solution for this but I'd like to get it merged so I can fix some other issues for the  selection user story. I'll merge this and create a bug for the button focusout

PR #92, opened 2019-06-28
Feature pdf annotations/em 1716 1345 delete highlight

benouaer:
https://tools.hmcts.net/jira/browse/EM-1716
https://tools.hmcts.net/jira/browse/EM-1345
### Change description ###
enable deletion of highlights/rectangles from the frontend, propagating the change through to the backend, as well as deleting the annotation if no more highlights exist
```
[ ] Yes
[X] No
```

### DIFF
--- a/proxy.config.js
+++ b/proxy.config.js
@@ -1,0 +1,23 @@
+module.exports = [
+  {
+    context: [
+      '/documents'
+    ],
+    target: 'http://localhost:1337',
+    secure: false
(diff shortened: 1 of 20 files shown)


proxy.config.js
```diff
@@ -0,0 +1,23 @@
+module.exports = [
```
linusnorton: Not sure I understand why you've changed this

PR #93, opened 2019-06-28
various fixes after user story QA

linusnorton:
- Disable dragging while in draw mode
- Switch out of draw mode after creating a new highlight
- Correct update the API after moving or resizing a highlight
- Do not  create a highlight if it is less than 5px x 5px
- Fix deleting annotations


projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -38,25 +37,29 @@ export class AnnotationSetComponent {
   }
 
   public onAnnotationUpdate(annotation: Annotation) {
-    const annotations = this.annotationSet.annotations.filter(a => a.id !== annotation.id);
-
-    annotations.push(annotation);
+    this.api
+      .postAnnotation(annotation)
+      .subscribe(newAnnotation => {
+        const index = this.annotationSet.annotations.findIndex(a => a.id === newAnnotation.id);
```
benouaer: these state updates feel like they should be done in the service as a result of the api call

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -38,25 +37,29 @@ export class AnnotationSetComponent {
...
+      });
   }
 
   public onAnnotationDelete(annotation: Annotation) {
-    this.annotationSet.annotations = this.annotationSet.annotations.filter(a => a.id !== annotation.id);
-    this.api.postAnnotationSet(this.annotationSet);
+    this.api
+      .deleteAnnotation(annotation.id)
+      .subscribe(() => {
+        this.annotationSet.annotations = this.annotationSet.annotations.filter(a => a.id !== annotation.id);
```
benouaer: same comment as above in terms updating the state in the service post the api call..

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.html
```diff
@@ -6,7 +6,7 @@
   >
     <img #img *ngIf="url" [ngClass]="'rot' + rotation" data-hook="dm.viewer.img" [src]="url" [style.width]="img.naturalWidth * zoom + 'px'" (error)="onLoadError()"/>
     <mv-annotation-set [annotationSet]="annotationSet"
-                       [drawMode]="drawMode"
+                       [drawMode]="toolbarEventsService.drawMode"
```
benouaer: could we extract this property within the component rather than the view. it seems like the view doesn't need to know about the service, just the draw mode

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.ts
```diff
@@ -21,41 +21,24 @@ import {ToolbarEventsService} from '../../shared/toolbar-events.service';
...
   rotation = 0;
   zoom = 1;
 
-  // local array of any subscriptions so that we can tidy them up later
-  private subscriptions: Subscription[] = [];
-
   constructor(
     private readonly printService: PrintService,
-    private readonly toolbarEventsService: ToolbarEventsService
+    public readonly toolbarEventsService: ToolbarEventsService
```
benouaer: the suggested view change in the comment above would mean that this can stay private

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
linusnorton: I'm not sure I agree, the API client is dumb it shouldn't really do anything apart from handle comms with the API.

linusnorton: After discussion it was agreed the best course of action is to keep the API class as an ApiClient and create a Service that manages the state updates of an AnnotationSet. I'll create a refactor ticket for that.

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -18,9 +19,7 @@ export class AnnotationSetComponent {
   @Input() width: number;
   @Input() height: number;
   @Input() page: number;
-  @Input() drawMode: boolean;
-
-  @Output() update = new EventEmitter<Annotation>();
+  @Input() drawMode: BehaviorSubject<boolean>;
```
russellwardink: in light of your recent refactor, can you review how this is injected in?

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
russellwardink: potential orphan subscription. might want an ngOnDestroy to tidy these up

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
russellwardink: potential orphan subscription. might want an ngOnDestroy to tidy these up

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -95,9 +98,13 @@ export class AnnotationSetComponent {
         type: 'highlight'
       };
 
-      this.api
-        .postAnnotation(annotation)
-        .subscribe(a => this.annotationSet.annotations.push(a));
+      if (rectangle.height > 5 || rectangle.width > 5) {
+        this.api
+          .postAnnotation(annotation)
+          .subscribe(a => this.annotationSet.annotations.push(a));
```
russellwardink: potential orphan subscription. might want an ngOnDestroy to tidy these up

PR #95, opened 2019-07-01
Toolbar button visibility service

linusnorton:
- Move toolbar button visibility state into service and use DI 
- Move toolbar button state into ToolbarModule
- Nest toolbar components inside main toolbar


projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  public showZoom = false;
+  public showRotate = false;
+  public showPresentationMode = false;
+  public showOpenFile = false;
+  public showBookmark = false;
+  public showHighlight = false;
+  public showSearchBarToggle = false;
+  public showSubToolbarToggle = false;
+  public showSidebarToggle = false;
+
```
russellwardink: Feels like the Annotation has been missed?

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
```diff
@@ -6,7 +6,7 @@ export type HighlightMode = boolean;
 export type DrawMode = boolean;
 
 @Injectable({providedIn: 'root'})
```
russellwardink: As you’ve moved these as Providers in the Toolbar Module, providedIn: root is a bit conflicting, as it will no longer use the root injector. Might be worth removing.

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
```diff
@@ -0,0 +1,71 @@
+import { BehaviorSubject } from 'rxjs';
+
```
russellwardink: Missing the @Injectable annotation

tslint.json
```diff
@@ -15,7 +15,7 @@
       "severity": "warn"
     },
     "eofline": true,
-    "forin": true,
+    "forin": false,
```
russellwardink: why this change?

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
linusnorton: These are the toolbar buttons rather than the toggles at the top of the page to turn features on/off

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
linusnorton: Are you sure? I thought it was natural for multiple classes to use the root injector... I will google it

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
linusnorton: I noticed in the second PR :)

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
russellwardink: should the draw and highlight mode not be included then?

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
russellwardink: tbh, not 100% but i believe the providedIn is a short-cut… not sure which one would take precedence.

projects/media-viewer/src/lib/media-viewer.component.html
```diff
@@ -30,8 +28,7 @@
                    [rotateOperation]="actionEvents?.rotate | async"
                    [setCurrentPage]="actionEvents?.setCurrentPage | async"
                    [changePageByDelta]="actionEvents?.changePageByDelta | async"
-                   [currentPageChanged]="currentPageChanged"
-                   [searchBarHidden]="toolbarButtonToggles?.searchBarHidden | async">
```
velimd: why was this removed?

tslint.json
linusnorton: Long, long ago in a JavaScript far away... libraries would modify the prototype of javascript objects in order to paper over the shoddy standard library. For example, it's entirely possible to set `Object.prototype.foo = () => console.log('foo');` and every object will then have a foo function. Potentially this means using a for in can be dangerous as you end up iterating keys you didn't expect. This linting rule checks that you have a `.hasOwnProperty` on the key. I've disabled it as it's not something that is done any more.

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
linusnorton: yes, they should. I'm not sure we have added the ability to turn them on/off

russellwardink: at present, i had them enabled under the wierd top toolbar Annotate switch, but they should in theory follow the same convention now outlined above

tslint.json
russellwardink: okay cool

projects/media-viewer/src/lib/viewers/viewer-event.service.spec.ts
```diff
@@ -1,12 +1,12 @@
-import {ViewerEventsService} from './viewer-events.service';
+import {ViewerEventService} from './viewer-event.service';
 import {Subscription} from 'rxjs';
```
velimd: spacing

projects/media-viewer/src/lib/media-viewer.component.html
linusnorton: uses dependency injection now

projects/media-viewer/src/lib/media-viewer.component.html
```diff
@@ -1,4 +1,4 @@
-<div id="outerContainer" [ngClass]="{ sidebarOpen: toolbarButtonToggles?.sidebarOpen | async }">
+<div id="outerContainer" [ngClass]="{ sidebarOpen: toolbarButtons.sidebarOpen | async }">
```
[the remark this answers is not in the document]
linusnorton: nope

projects/media-viewer/src/lib/viewers/viewer-event.service.spec.ts
linusnorton: I didn't want to make unnecessary formatting changes as part of this PR.... it's already big enough

PR #100, opened 2019-07-03
Feature pdf annotations/em 1748 popup toolbar comment

benouaer:
https://tools.hmcts.net/jira/browse/EM-1717
### Change description ###
Enable creation of comments
```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
+++ b/projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
@@ -16,16 +16,17 @@ export class CommentComponent {
   fullComment: string;
   author: User;
   editor: User;
+  _editable: boolean;
+  _selected: boolean;
 
   @Output() click = new EventEmitter();
   @Output() delete = new EventEmitter();
   @Output() updated = new EventEmitter<String>();
-  @Input() selected: boolean;
-  @Input() editable: boolean;
   @Input() rotate = 0;
   @Input() zoom = 1;
   @Input() rectangle: Rectangle;
   @ViewChild('form') form: ElementRef;
+  @ViewChild('textArea') textArea: ElementRef;
 
   @Input()
   set comment(comment: Comment) {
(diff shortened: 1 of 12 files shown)


projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
```diff
@@ -12,20 +12,21 @@ export class CommentComponent {
 
   private readonly MAX_COMMENT_LENGTH = 50;
 
-  editable = false;
   lastUpdate: string;
   fullComment: string;
   author: User;
   editor: User;
+  _editable: boolean;
```
linusnorton: eh? we've never used _ notation before... any reason to start now?

projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
```diff
@@ -35,6 +36,34 @@ export class CommentComponent {
     this.fullComment = comment.content;
   }
 
+  @Input()
+  set selected(selected: boolean) {
```
linusnorton: we'll this is confusing

projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
```diff
@@ -35,6 +36,34 @@ export class CommentComponent {
...
+    this._selected = selected;
+  }
+
+  get selected() {
+    return this._selected || Boolean(!this.author && this.fullComment);
+  }
+
+  @Input()
+  set editable(editable: boolean) {
```
linusnorton: I'm confused, why are we using set on these?

projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
benouaer: it's so that we can take into account the logic for draft comments, so that they remain selected until they press save

projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
benouaer: same as above, to handle draft comments

projects/media-viewer/src/lib/annotations/annotation-set/annotation/comment/comment.component.ts
benouaer: it's because I had to introduce the setters/getters for the reasons explained below

PR #101, opened 2019-07-03
Phani/e2e cucumberr

phaniperla:

### Change description ###

```
[ ] Yes
[ ] No
```


e2e/cucumber.conf.js
```diff
@@ -13,7 +12,12 @@ exports.config = {
   specs: [
     './src/**/*.feature',
   ],
-  onPrepare() {
+  onPrepare: function () {
+
+    const {Given, Then, When} = require('cucumber');
+    global.Given = Given;
+    global.When = When;
+    global.Then = Then;
```
linusnorton: this is not necessary

e2e/src/pages/app.po.ts
```diff
@@ -1,11 +1,11 @@
-import { browser, by, element, Locator, protractor } from 'protractor';
+import {browser, by, element, ElementFinder, Locator, protractor} from 'protractor';
 import { By } from '@angular/platform-browser';
 const until = protractor.ExpectedConditions;
 
 export class AppPage {
 
   async navigateTo() {
-    await browser.waitForAngularEnabled(false);
+    // await browser.waitForAngularEnabled(false);
```
linusnorton: why?

e2e/src/pages/app.po.ts
```diff
@@ -34,6 +34,8 @@ export class AppPage {
         this.clickElement(by.css('label[for="zoom-btn-toggle"]')),
         this.clickElement(by.css('label[for="toggleAnnotations"]'))
       ]);
+    } else {
+      console.log('allready checked');
```
linusnorton: remove this please

e2e/src/pages/app.po.ts
```diff
@@ -54,12 +56,26 @@ export class AppPage {
...
     await browser.wait(async () => {
       return (await element.all(selector).isPresent());
-    }, 10000, 'failed to load search results');
+    }, 30000, 'failed to load search results');
+  }
+
+  async getItem(el: ElementFinder) {
+    const item = protractor.promise.defer();
+    browser.wait(function() {
+      return element(by.id('el')).getAttribute('attribute').then(function(value) {
```
linusnorton: please don't use .then use await

e2e/src/pages/app.po.ts
```diff
@@ -54,12 +56,26 @@ export class AppPage {
...
   async waitForElementsArray(selector: Locator) {
     await browser.wait(async () => {
       return (await element.all(selector).isPresent());
-    }, 10000, 'failed to load search results');
+    }, 30000, 'failed to load search results');
+  }
+
+  async getItem(el: ElementFinder) {
+    const item = protractor.promise.defer();
+    browser.wait(function() {
```
linusnorton: please use shorthand notation: `() => {`

e2e/src/pages/navigate.po.ts
```diff
@@ -1,20 +1,32 @@
...
+    this.pageNumber = element(by.id('pageNumber'));
   }
 
   async number() {
     await this.waitForElement(by.id('pageNumber'));
     return element(by.id('pageNumber')).getAttribute('value');
   }
+
+  async goToNextPage() {
+    await this.next.click().then(async function () {
```
linusnorton: I'm confused by this, you're mixing .then and await?

e2e/src/pages/navigate.po.ts
```diff
@@ -1,20 +1,32 @@
...
+      await browser.driver.sleep(2000);
```
linusnorton: this should not be necessary

e2e/src/step-definitions/hooks.ts
```diff
@@ -1,13 +1,24 @@
-import { After, Status } from 'cucumber';
-import { browser } from 'protractor';
+import {After, Before, Status} from 'cucumber';
+import {browser} from 'protractor';
 import * as fs from 'fs';
 
-After(async function(scenario) {
+Before({tags: '@MediaViewer'}, async () => {
+  await browser.waitForAngularEnabled(false);
+  return await browser.driver.navigate().to(browser.baseUrl);
```
linusnorton: what's this return statement for?

e2e/src/step-definitions/hooks.ts
```diff
@@ -1,13 +1,24 @@
...
+});
+
+After(function () {
+  console.log('Successfully run the tests');
```
linusnorton: This seems unnecessary

e2e/src/step-definitions/hooks.ts
```diff
@@ -1,13 +1,24 @@
...
+
+});
+
+After(async (scenario) => {
   if (scenario.result.status === Status.FAILED) {
     const screenshot = await browser.takeScreenshot();
-    const stream = fs.createWriteStream('./reports/fail.png');
-
+    const stream = fs.createWriteStream('./fail.png');
```
linusnorton: this should be in the reports folder no?

e2e/src/features/mediaviewer.feature
```diff
@@ -0,0 +1,30 @@
...
+    And I click next button on the pdf
+    Then I should see next page number should be '<page_number>'
+    And I click previous button on the pdf
+    Then I should see previous page number should be '<page_number2>'
+
+    Examples:
+      | page_number |page_number2|
+      | 2           |       1     |
+
+
```
benouaer: why does this not follow the cucumber pattern of Given/When/Then?

e2e/cucumber.conf.js
phaniperla: Agree with you ...  i thought I want call those cucumber calls globally  so am gonna remove that code

e2e/src/pages/app.po.ts
phaniperla: That's been added for debugging purpose and will delete that code

e2e/src/pages/navigate.po.ts
```diff
@@ -1,20 +1,32 @@
...
+    });
+  }
+
+  async goToPreviousPage() {
+    await this.previous.click().then(async function () {
+      await browser.driver.sleep(2000);
+    });
+  }
```
benouaer: I don't get why this is being changed?
also, I would avoid using the sleep method if possible, as the time to wait changes depending on the environment
there are other mechanisms like waitForElement that we're using elsewhere (well, in the previous code)

e2e/src/pages/app.po.ts
phaniperla: I tried different ways to get the value from getAttribute() method but some reason that's not working that's why i used old fashion of using promise

phaniperla: Let's discuss this

e2e/src/pages/app.po.ts
phaniperla: Agree with you...

e2e/src/step-definitions/hooks.ts
phaniperla: that's return driver object... let's discuss this...

e2e/src/step-definitions/media-viewer.ts
```diff
@@ -1,24 +1,53 @@
...
 
 const page = new AppPage();
+const navigatePage: NavigatePage = new NavigatePage();
 
-Given('I am on Media Viewer Page',  async () => {
+Given('I am on Media Viewer Page', async () => {
   await page.preparePage();
 });
 
 Given('I enable toggle buttons', async () => {
```
benouaer: this seems like an action, so shouldn't it be When(....)

e2e/src/step-definitions/media-viewer.ts
```diff
@@ -1,24 +1,53 @@
...
+
+When('I click previous button on the pdf', async () => {
+  await page.selectPdfViewer();
+  await page.preparePage();
+  await page.waitForPdfToLoad();
+  await navigatePage.goToPreviousPage();
+});
+
+Then('I should see next page number should be {string}', async (text: string) => {
+  navigatePage.pageNumber.getAttribute('value').then(function (value) {
```
benouaer: it would be consistent with the rest of the code to use await instead of then

e2e/src/step-definitions/media-viewer.ts
```diff
@@ -1,24 +1,53 @@
...
+    chai.expect(value).to.equal(text);
+  });
+});
+
+
+Then('I should see previous page number should be {string}', async (text: string) => {
+  navigatePage.pageNumber.getAttribute('value').then(function (value) {
```
benouaer: same as above about use of await instead of then

e2e/src/pages/navigate.po.ts
phaniperla: fixed

e2e/src/pages/navigate.po.ts
phaniperla: fixed

e2e/src/step-definitions/media-viewer.ts
phaniperla: fixed

e2e/src/step-definitions/media-viewer.ts
phaniperla: fixed

phaniperla: getAttribute() method not returning the value that's the reason i used ES5 syntax which is working perfectly fine. 
looks like that's a protractor issue

e2e/src/step-definitions/media-viewer.ts
phaniperla: fixed

e2e/src/pages/navigate.po.ts
phaniperla: removed ES5 promise syntax and used ES7 sysntax

e2e/src/features/mediaviewer.feature
phaniperla: fixed

e2e/protractor.conf.js
```diff
@@ -12,7 +12,7 @@ exports.config = {
 // Firefox local setup
   exclude: ['./src/**/download.e2e-spec.ts', "./src/**/print.e2e-spec.ts"],
   capabilities: {
-    browserName: 'firefox',
+    browserName: 'chrome',
```
JeroenRijks: could you uncomment the Chrome local setup section and comment the Firefox local setup bit out? That way people won't have to look into the code to switch between the two.

e2e/src/pages/navigate.po.ts
```diff
@@ -1,20 +1,28 @@
...
-    await this.waitForElement(by.id('next'));
-    await this.clickElement(by.id('next'));
-  }
-
-  async goToPreviousPage() {
-    await this.waitForElement(by.id('previous'));
-    await this.clickElement(by.id('previous'));
+  constructor() {
+    super();
+    this.next = element(by.id('next'));
```
JeroenRijks: this.next is set to a promise. when you do goToNextPage(), you're going to be clicking on a promise, without knowing if it's resolved or not. That means it will sometimes work, and sometimes not work.
to fix, you could set
this.next = await element(by.id('next'));
and do something similar for all other elements that you're defining

e2e/src/step-definitions/media-viewer.ts
```diff
@@ -1,24 +1,51 @@
...
+  await page.selectPdfViewer();
+  await page.preparePage();
+  await page.waitForPdfToLoad();
+  await navigatePage.goToPreviousPage();
+});
 
-  chai.expect(header).to.equal(text);
+Then('I should see next page number should be {string}', async (text: string) => {
+  const value = await navigatePage.pageNumber.getAttribute('value');
+  console.log('PageNumber99::->' + value);
```
linusnorton: can we remove the console log

e2e/protractor.conf.js
phaniperla: cucumber e2e tests using cucumber.conf.js

e2e/src/pages/navigate.po.ts
phaniperla: Thanks @JeroenRijks  I will incorporate your review comments

phaniperla: I cannot declare constructor() as async as the compiler throwing an error message

e2e/src/step-definitions/media-viewer.ts
phaniperla: Done

PR #127, opened 2019-08-14
Fix/toolbar visibility

velimd:
### Change description ###

```
[ ] Yes
[ ] No
```


projects/media-viewer/src/public_api.ts
```diff
@@ -7,3 +7,4 @@ export * from './lib/toolbar/toolbar.module';
 export * from './lib/comments-summary/comments-summary.module';
 export * from './lib/annotations/annotation-api.service';
 export * from './lib/annotations/annotations.module';
+export * from './lib/toolbar/toolbar.module';
```
russellwardink: This is already present on line 6....

velimd: amended

PR #153, opened 2019-09-05
Em 1387 document outline

velimd:

### Change description ###

```
[ ] Yes
[ ] No
```


projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    expect(component.renderOutlineView).toHaveBeenCalled();
+  });
+
+  it('should call renderOutlineView and set the outline to empty when a new url is loaded', () => {
+    const mockDocument = {
+      getOutline: () => new Promise(resolve => {
+        resolve({});
+      })
+    };
+    const pdf = new Promise(resolve => {
```
linusnorton: `const pdf = Promise.resolve(mockDocument);`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      url: new SimpleChange('a', 'b', true)
+    });
+
+    expect(component.renderOutlineView).toHaveBeenCalled();
+  });
+
+  it('should get the document outline and render it', async () => {
+    spyOn(component, 'render');
+    const mockDocument = {
+      getOutline: () => new Promise(resolve => {
```
linusnorton: See previous note on `Promise.resolve`

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      cMapUrl: 'assets/minified/cmaps',
+      cMapPacked: true,
+      withCredentials: true
+    });
+
+    this.outline = await pdf.getOutline();
+    this.render({ outline: this.outline });
+  }
+
+  bindOutlineLink(element, { url, newWindow, dest }) {
```
linusnorton: Proper type info here would be nice, wrap that second argument into an interface perhaps

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+    const linkService = this.pdfDocumentOutline.linkService;
+
+    element.href = linkService.getDestinationHash(dest);
+    element.onclick = () => {
```
linusnorton: why are we doing this sort of stuff, it doesn't feel very angular?

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      if (dest) {
+        linkService.navigateTo(dest);
+      }
+      return false;
+    };
+  }
+
+  setStyles(element, { bold, italic }) {
```
linusnorton: again this feels a bit weird

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  resetOutline() {
+    let firstNode = this.outlineContainer.nativeElement.firstElementChild;
+    while (firstNode) {
+      this.outlineContainer.nativeElement.removeChild(firstNode);
+      firstNode = this.outlineContainer.nativeElement.firstElementChild;
+    }
+  }
+
+  render({ outline, }) {
```
linusnorton: trailing comma seems redundant and type info would be nice

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+    let outlineCount = 0;
+
+    if (this.outline) {
+      this.resetOutline();
+    }
+    this.outline = outline || null;
```
linusnorton: what values does outline have? is null more useful that false or undefined?

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      for (const item in levelData.items) {
+        const div = document.createElement('div');
+        div.className = 'outlineItem';
+
+        const element = document.createElement('a');
+        this.bindOutlineLink(element, levelData.items[item]);
+        this.setStyles(element, levelData.items[item]);
+        element.textContent = pdfjsLib.removeNullCharacters(levelData.items[item].title) || 'outline ' + outlineCount + 1;
+
+        div.appendChild(element);
```
linusnorton: I think lines 109-117 can be extracted into another method, and perhaps should even be extracted into a new component for OutlineItem or something

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (this.outline) {
+      this.resetOutline();
+    }
+    this.outline = outline || null;
+
+    const fragment = document.createDocumentFragment();
+    const queue = [{ parent: fragment, items: this.outline, }];
+    let hasAnyNesting = false;
+    while (queue.length > 0) {
```
linusnorton: Okay, so what we have here is a tree structure rather than a queue.

The best way to iterate a tree is through recursion, so I would split this loop out into it's own method that took an arbitrary point in the tree, created the dom element and then scanned for children. Everything above this line would stay as a bootstrap for the recursive function. For example:

```
interface TreeNode {
  title: string | null, // this seems to be null and default to outline + outlineCount + 1
  parent: TreeNode //or you could use Fragment if that's preferred
  children: TreeNode[]
}

function renderTree(node: TreeNode) {
  this.createTreeNodeDomElement(node.title);

  for (const child of node.children) {
    this.renderTree(child);
  }
}
```

Obviously this a vast simplification of what is happening but it should suffice to give you the idea. The outlineCount can be passed to the `renderTree` method if it is required.

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  async ngOnChanges(changes: SimpleChanges) {
+    if (changes.url) {
+      this.outline = [];
+      await this.renderOutlineView();
+    }
+  }
+
+  async renderOutlineView() {
+    const pdf = await pdfjsLib.getDocument({
```
linusnorton: Are you sure we should be fetching the entire document again? Can't we store it someone once it gets loaded?

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-js/pdf-js-wrapper.provider.ts
```diff
@@ -7,27 +8,30 @@ import { ToolbarEventService } from '../../../toolbar/toolbar-event.service';
...
 export class PdfJsWrapperFactory {
 
+  private linkService: pdfjsViewer.PDFLinkService;
+  private eventBus: pdfjsViewer.EventBus;
+
   constructor(
-    private readonly toolbarEvents: ToolbarEventService
-  ) {}
+    private readonly toolbarEvents: ToolbarEventService) {
+    this.linkService = new pdfjsViewer.PDFLinkService();
```
linusnorton: Why do we make a new pdfjsViewer for the link service and the event bus? Shouldn't they be from the same object?

projects/media-viewer/src/lib/toolbar/side-bar/outline-view/outline-view.component.ts
velimd: the issue is the sidebar and main viewer are loaded at the same time, so this becomes a problem with getting the document.

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-js/pdf-js-wrapper.provider.ts
velimd: it was previous created when the create function was called, we just moved it and stored it as a variable to be accessed in the outline component.

PR #155, opened 2019-09-06
Bug/em2022 detect annotation toggle changes

benouaer:
- https://tools.hmcts.net/jira/browse/EM-2022
- https://tools.hmcts.net/jira/browse/EM-2023

### Change description ###
- react to changes to annotation toggle
- enable overriding of toolbar buttons

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
+++ b/projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
@@ -11,7 +11,7 @@ import { SelectionAnnotation } from '../../annotation.service';
 })
 export class CommentComponent implements OnChanges {
 
-  readonly MAX_COMMENT_LENGTH = 48;
+  readonly MAX_COMMENT_LENGTH;
 
   lastUpdate: string;
   originalComment: string;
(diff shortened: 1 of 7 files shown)


projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -35,6 +35,10 @@ export class CommentComponent implements OnChanges {
   @ViewChild('form') form: ElementRef;
   @ViewChild('textArea') textArea: ElementRef;
 
+  constructor() {
+    this.MAX_COMMENT_LENGTH = 48;
```
velimd: is this best practise, just want to know

projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -71,23 +71,40 @@ export class MediaViewerComponent implements OnChanges, AfterContentInit {
...
     }
   }
 
   onLoadException(exception: ViewerException) {
     this.viewerException.emit(exception);
   }
 }
+
+enum SupportedContentTypes {
+  PDF = 'pdf',
```
velimd: nice, i was always wondering why we never use an enum

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
benouaer: the other way is not supported in typescript <3.1.0, which JUI use

linusnorton:
Does it matter? Aren't they consuming compiled JS instead of raw TS?

PR #171, opened 2019-09-17
EM-1927: Implement 5000 char limit on comment box.

tomxelliott:
- add limit to html element
	- slice the string in the typescript code if for some reason it is not caught by the html limit

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
+++ b/projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
@@ -12,6 +12,7 @@ import { SelectionAnnotation } from '../../annotation.service';
 export class CommentComponent implements OnChanges {
 
   readonly MAX_COMMENT_LENGTH;
+  readonly COMMENT_CHAR_LIMIT;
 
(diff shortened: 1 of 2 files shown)


projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -95,7 +95,7 @@ export class CommentComponent implements OnChanges {
   }
 
   public onSave() {
-    this._comment.content = this.fullComment;
+    this._comment.content = this.fullComment.length > 5000 ? this.fullComment.slice(0, 5000) : this.fullComment;
```
linusnorton: this can just be `this.fullComment.substring(0, 5000);` if the string is less than 5000 characters substring and slice will use the max string length. I'd say use substring as it's more intuitive to what is happening.

linusnorton: Also, maybe that value of 5000 should be put into a const somewhere so it doesn't need to be changed twice

PR #186, opened 2019-11-06
EM-2102: Add label to the page number input element

tomxelliott:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
+++ b/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
@@ -20,10 +20,12 @@
         <span data-l10n-id="next_label">Next</span>
       </button>
     </div>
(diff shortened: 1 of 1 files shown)


projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
```diff
@@ -20,10 +20,12 @@
         <span data-l10n-id="next_label">Next</span>
       </button>
     </div>
-    <input type="number" id="pageNumber" class="toolbarField pageNumber" title="Page" value="1" size="4" min="1"
```
benouaer: I thought this just needed an aria-label tag for accessbility

tomxelliott: yep, changing it now. I was trying to use the recommended approach which is to wrap them in labels but doesn't seem to work. Will be changing to aria-label tags now.

PR #195, opened 2019-11-11
init impl of key shortcut bindngs

pawelkaczmarekuk:
Example approach to key bindings

### DIFF
--- a/projects/media-viewer/src/lib/key-strokes/defult-key-bindings.json
+++ b/projects/media-viewer/src/lib/key-strokes/defult-key-bindings.json
@@ -1,0 +1,3 @@
+[
+  { "code": "ArrowDown", "ctrlKey": false, "altKey": true, "actionId": "NEXT_PAGE"}
+]
--- a/projects/media-viewer/src/lib/media-viewer.component.ts
+++ b/projects/media-viewer/src/lib/media-viewer.component.ts
@@ -1,7 +1,7 @@
 import {
   AfterContentInit,
   Component,
-  EventEmitter,
+  EventEmitter, HostListener,
   Input,
   OnChanges,
   Output,
@@ -18,6 +18,7 @@ import { AnnotationSet } from './annotations/annotation-set/annotation-set.model
 import { ToolbarEventService } from './toolbar/toolbar-event.service';
 import { AnnotationApiService } from './annotations/annotation-api.service';
 import { ResponseType, ViewerException } from './viewers/error-message/viewer-exception.model';
+import {KeyListenerService} from './key-strokes/key-listener.service';
 
 @Component({
   selector: 'mv-media-viewer',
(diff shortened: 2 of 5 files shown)


projects/media-viewer/src/lib/toolbar/main-toolbar/search-bar/search-bar.component.ts
```diff
@@ -40,11 +40,14 @@ export class SearchBarComponent implements OnInit, OnDestroy {
 
   @HostListener('window:keydown', ['$event'])
```
benouaer: should this code be moved from here then? into the key-listener-service

projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -47,6 +48,7 @@ export class MediaViewerComponent implements OnChanges, AfterContentInit {
   constructor(
     public readonly toolbarButtons: ToolbarButtonVisibilityService,
     public readonly toolbarEvents: ToolbarEventService,
+    public readonly keyListenerService: KeyListenerService,
```
benouaer: I'm wondering whether it would be better to use a directive for this, so that even the host-listener stuff can be contained within that? just a thought, not sure if it serves the purpose better, but what it would give us is the whole functionality contained within the directive, and we just tag the media-viewer component with it to enable it

projects/media-viewer/src/lib/key-strokes/defult-key-bindings.json
```diff
@@ -0,0 +1,3 @@
+[
+  { "code": "ArrowDown", "ctrlKey": false, "altKey": true, "actionId": "NEXT_PAGE"}
```
benouaer: do we need the shift key here as well?

projects/media-viewer/src/lib/toolbar/main-toolbar/search-bar/search-bar.component.ts
pawelkaczmarekuk: HostListeners must be a part of a component as far as I can tell.

projects/media-viewer/src/lib/key-strokes/defult-key-bindings.json
pawelkaczmarekuk: It well might be the case. It's just an example.

projects/media-viewer/src/lib/media-viewer.component.ts
pawelkaczmarekuk: It's worth a try.

pawelkaczmarekuk: The listener is attached to the top component. But perhaps not all the bindings should be handled globally - for example in case of a modal or something similar...

projects/media-viewer/src/lib/toolbar/main-toolbar/search-bar/search-bar.component.ts
benouaer: ah ok, but it seems like we're doing key binding and matching here, so should this delegate to the service like the media-viewer does?

pawelkaczmarekuk: Sry - this HostListener was not added by me. I believe this particular Listener could be removed after we add the global handling of key strokes.

PR #198, opened 2019-11-12
annotation buttons position scales horizontally and dissapears if can…

pawelkaczmarekuk:
'annotation buttons position scales horizontally and dissapears if cant fit when scalling windows horizontally or zooming by browser'

### DIFF
--- a/projects/media-viewer/src/lib/styles/media.scss
+++ b/projects/media-viewer/src/lib/styles/media.scss
@@ -360,7 +360,7 @@
 }
 
 @media all and (max-width: 1000px) {
-  .hiddenIfLargerThan1000px, .hiddenIfLargerThan1000px * {
+  div.hiddenIfLargerThan1000px, div.hiddenIfLargerThan1000px * {
     display: none;
   }
 }
--- a/projects/media-viewer/src/lib/styles/media.scss
+++ b/projects/media-viewer/src/lib/styles/media.scss
@@ -174,6 +174,14 @@
     content: url(/assets/images/secondaryToolbarButton-spreadEven@2x.png);
   }
 
+  .secondaryToolbarButton.highlightBtn::before {
+    content: url(/assets/images/toolbarButton-highlight@2x.png);
+  }
+
+  .secondaryToolbarButton.drawBtn::before {
+    content: url(/assets/images/toggleButton-draw@2x.png);
+  }
+
   .toolbarButton.highlightBtn::before {
     content: url(/assets/images/toolbarButton-highlight@2x.png);
   }
--- a/projects/media-viewer/src/lib/styles/media.scss
+++ b/projects/media-viewer/src/lib/styles/media.scss
@@ -352,7 +352,7 @@
 }
 
 @media all and (max-width: 1000px) {
-  .hidden1000px, .hidden1000px * {
+  .hiddenIfLargerThan1000px, .hiddenIfLargerThan1000px * {
     display: none;
   }
 }
--- a/projects/media-viewer/src/lib/styles/media.scss
+++ b/projects/media-viewer/src/lib/styles/media.scss
@@ -292,37 +292,37 @@
     background-color: hsla(0,0%,0%,.7);
   }
 
-  #outerContainer.sidebarOpen #viewerContainer {
+  sidebarOpen #viewerContainer {
     left: 0px !important;
   }
-  html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer {
+  html[dir='rtl'] sidebarOpen #viewerContainer {
     right: 0px !important;
   }
 
-  #outerContainer .hiddenLargeView,
-  #outerContainer .hiddenMediumView {
+  .hiddenLargeView,
+  .hiddenMediumView {
     display: inherit;
   }
-  #outerContainer .visibleLargeView,
-  #outerContainer .visibleMediumView {
+  .visibleLargeView,
+  .visibleMediumView {
     display: none;
   }
 }
 
 @media all and (max-width: 770px) {
-  #outerContainer .hiddenLargeView {
+  .hiddenLargeView {
     display: none;
   }
-  #outerContainer .visibleLargeView {
+  .visibleLargeView {
     display: inherit;
   }
 }
 
 @media all and (max-width: 700px) {
-  #outerContainer .hiddenMediumView {
+  .hiddenMediumView {
     display: none;
   }
-  #outerContainer .visibleMediumView {
+  .visibleMediumView {
     display: inherit;
   }
 }
@@ -349,4 +349,16 @@
   #scaleSelectContainer {
     display: none;
   }
+}
+
+@media all and (max-width: 1000px) {
+  .hidden1000px, .hidden1000px * {
+    display: none;
+  }
+}
+
+#annotationButtons {
+  position: absolute;
+  left: 23%;
+  transform: translateX(-23%);
 }
--- a/projects/media-viewer/src/lib/styles/media.scss
+++ b/projects/media-viewer/src/lib/styles/media.scss
@@ -292,10 +292,10 @@
     background-color: hsla(0,0%,0%,.7);
   }
 
-  sidebarOpen #viewerContainer {
+  .sidebarOpen #viewerContainer {
     left: 0px !important;
   }
-  html[dir='rtl'] sidebarOpen #viewerContainer {
+  html[dir='rtl'] .sidebarOpen #viewerContainer {
     right: 0px !important;
   }
 
--- a/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
+++ b/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
@@ -25,6 +25,7 @@
          tabindex="0" data-l10n-id="page" (change)="onPageNumberInputChange(pageNumberInput.value)" #pageNumberInput>
     <span id="numPages" class="toolbarLabel"></span>
   </ng-container>
+  <div class="toolbarButtonSpacer"></div>
 </div>
 <div id="annotationButtons">
   <div class="splitToolbarButton hiddenIfLargerThan1000px" >
--- a/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
+++ b/projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
@@ -25,11 +25,15 @@
          tabindex="0" data-l10n-id="page" (change)="onPageNumberInputChange(pageNumberInput.value)" #pageNumberInput>
     <span id="numPages" class="toolbarLabel"></span>
   </ng-container>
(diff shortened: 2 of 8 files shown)


jai-santhosh:
Need some clarification from PO as this will disable user from creating annotation. Highlight and draw buttons are hidden from 200% zoom onwards.

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -292,37 +292,37 @@
     background-color: hsla(0,0%,0%,.7);
   }
 
-  #outerContainer.sidebarOpen #viewerContainer {
+  sidebarOpen #viewerContainer {
```
velimd: i believe you still need the dot ".sideOpen" in order to identify the class name

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -292,37 +292,37 @@
...
     left: 0px !important;
   }
-  html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer {
+  html[dir='rtl'] sidebarOpen #viewerContainer {
```
velimd: same with the dot "."

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
```diff
@@ -25,11 +25,15 @@
...
-  <button *ngIf="toolbarButtons.showHighlightButton" (click)="onClickHighlightToggle()" [class.toggled]="toolbarEvents.highlightMode | async" class="toolbarButton highlightBtn" title="Highlight Text Mode" id="toggleHighlightButton" tabindex="0" data-l10n-id="toggleHighlightButton">
-    <span data-l10n-id="next_label">Highlight Text</span>
-  </button>
-  <button (click)="onClickDrawToggle()" *ngIf="toolbarButtons.showDrawButton" [class.toggled]="toolbarEvents.drawMode | async" class="toolbarButton drawBtn" title="Draw Mode" id="toggleDrawButton" tabindex="0" data-l10n-id="toggleDrawButton">
-    <span data-l10n-id="next_label">Draw</span>
-  </button>
+
+</div>
+<div id="annotationButtons">
+  <div class="splitToolbarButton hidden1000px" >
```
velimd: this is just nitpicking but maybe change the name as 1000px isnt the best thing.

projects/media-viewer/src/lib/styles/media.scss
pawelkaczmarekuk: nice catch

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
pawelkaczmarekuk: there already are names like hiddenLargeView, hiddenMediumView, ... etc that's why I wanted to be specific

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -292,37 +292,37 @@
     background-color: hsla(0,0%,0%,.7);
   }
 
-  #outerContainer.sidebarOpen #viewerContainer {
```
benouaer: is this not needed anymore?

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
benouaer: it's not clear from the name whether it's hidden for width higher or lower than 1000px

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -350,3 +350,15 @@
     display: none;
   }
 }
+
+@media all and (max-width: 1000px) {
+  .hidden1000px, .hidden1000px * {
```
benouaer: why 1000px?

pawelkaczmarekuk: it turns out to be the right width for the annotation buttons to disappear

projects/media-viewer/src/lib/styles/media.scss
pawelkaczmarekuk: fixed

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
pawelkaczmarekuk: I will rename

projects/media-viewer/src/lib/styles/media.scss
pawelkaczmarekuk: nope

projects/media-viewer/src/lib/toolbar/main-toolbar/right-pane/right-pane.component.html
```diff
@@ -25,8 +25,7 @@
     <span data-l10n-id="bookmark_label">Current View</span>
   </a>
   <div class="toolbarButtonSpacer"></div>
-  <button *ngIf="toolbarButtons.showSubToolbar"
-          id="secondaryToolbarToggle" class="toolbarButton" title="Tools" tabindex="0" data-l10n-id="tools"
+  <button id="secondaryToolbarToggle" class="toolbarButton" title="Tools" tabindex="0" data-l10n-id="tools"
```
velimd: there needs to be a condition so when the other buttons disappear this is placed, otherwise we have duplicated buttons.

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
```diff
@@ -25,11 +25,14 @@
          tabindex="0" data-l10n-id="page" (change)="onPageNumberInputChange(pageNumberInput.value)" #pageNumberInput>
     <span id="numPages" class="toolbarLabel"></span>
   </ng-container>
-  <div class="toolbarButtonSpacer"></div>
```
velimd: might have been a reason why this is here, i dont remember but i know its in other parts of the toolbar html (mostly likely styles).

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.ts
```diff
@@ -14,12 +14,40 @@ export class SubToolbarComponent {
     public readonly toolbarEvents: ToolbarEventService
   ) {}
 
+  onClickHighlightToggle() {
+    this.toolbarEvents.highlightMode.next(true);
```
velimd: i believe we already have a function in toolbar events that does this. so if hightlight was selected, draw is set to false.

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.ts
```diff
@@ -14,12 +14,40 @@ export class SubToolbarComponent {
...
+    this.toolbarEvents.drawMode.next(false);
+    this.closeMenu();
+  }
+
+  onClickDrawToggle() {
+    this.toolbarEvents.drawMode.next(true);
```
velimd: same as above

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -350,3 +358,15 @@
     display: none;
   }
 }
+
+@media all and (max-width: 1000px) {
+  .hiddenIfLargerThan1000px, .hiddenIfLargerThan1000px * {
```
velimd: im not sure the * works on class names. I know it works on element tags, but im sure it doesnt for class name.

projects/media-viewer/src/lib/toolbar/main-toolbar/right-pane/right-pane.component.html
pawelkaczmarekuk: that's what PO wanted:

pawel-k 11:55 AM
What buttons should go into submenu unfolded by the chevron button?
highlight
draw
rotate left
rotate right
? those buttons disappear when scalling
should chevron button be visible all the time?
Linda Souto Maior 11:59 AM
Ideally
Chevron button is only available when buttons are hidden
It makes sense to me to hide print and download first
and then maybe highlight and draw
it's odd though because highlight and draw appear before zoom and rotate, so from UX point doesn't make 100% sense
but then rotate/zoom are probably more used features by many users
or majority of users
pawel-k 12:01 PM
the sequence of buttons disappearing may be difficult to implement (edited) 
the fact that print button disappears first does not make more room for hightlight button
Linda Souto Maior 12:01 PM
Ah ok
then i suggest we move from right to left
if sequencing isn't possible
and then as less of the toolbar is visible, buttons are hidden from right-to-left
pawel-k 12:03 PM
ok - that's what happening in my branch atm. I would suggest chevron to be there all the time - otherwise some people may miss it when it appears. if it's there then people may try it before they need it...
Lawrie:racing_motorcycle: 12:10 PM
We're doing this more for accessibility to pick up the buttons and our toolbar is tactical so the chevron button could be there all the time?
Linda Souto Maior 12:14 PM
Yep, agreed!

projects/media-viewer/src/lib/toolbar/main-toolbar/left-pane/left-pane.component.html
pawelkaczmarekuk: I'll re-add.

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.ts
pawelkaczmarekuk: will fix

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.ts
pawelkaczmarekuk: will fix

projects/media-viewer/src/lib/styles/media.scss
pawelkaczmarekuk: will add div.

projects/media-viewer/src/lib/toolbar/main-toolbar/right-pane/right-pane.component.html
velimd: i guess is the PO wants it, i cant argue but i think its not nice to have buttons duplicated.

projects/media-viewer/src/lib/styles/media.scss
```diff
@@ -350,3 +358,15 @@
     display: none;
   }
 }
+
+@media all and (max-width: 1000px) {
+  div.hiddenIfLargerThan1000px, div.hiddenIfLargerThan1000px * {
```
velimd: just to confirm ive tested it and it works with class names. sorry about my previous comment.

PR #206, opened 2019-11-22
EM-1947: Emit state if comment has been updated but not saved

tomxelliott:
- show banner in wrapper if changes not saved
	- unsaved state is emitted from comment all the way up the chain for our consuming services to decide how to handle unsaved comments
	- tests updated

### Change description ###

```
[ ] Yes
[ ] No
```


projects/media-viewer/src/lib/annotations/comment-set/comment-set.component.ts
```diff
@@ -22,6 +25,8 @@ export class CommentSetComponent implements OnInit, OnDestroy {
   @Input() rotate: number;
   @Input() height: number;
 
+  @Output() unsavedChanges = new EventEmitter<boolean>();
```
linusnorton: Wouldn't a subject be a more angular way?

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -88,8 +89,17 @@ export class CommentComponent implements OnChanges {
     this.editable = true;
   }
 
+  onCommentChange(updatedComment) {
+    if (this.originalComment.substring(0, this.COMMENT_CHAR_LIMIT) !== updatedComment.substring(0, this.COMMENT_CHAR_LIMIT)) {
```
linusnorton:
```const hasUnsavedChanges = this.originalComment.substring(0, this.COMMENT_CHAR_LIMIT) !== updatedComment.substring(0, this.COMMENT_CHAR_LIMIT);
this.unsavedChanges.emit(hasUnsavedChanges);```

File: src/app/media-viewer-wrapper/media-viewer-wrapper.component.ts [hunk 3]
```diff
@@ -72,6 +73,11 @@ export class MediaViewerWrapperComponent implements AfterContentInit {
     setTimeout(() => this.mediaLoadStatus = undefined, 2000);
   }
 
+  onUnsavedChanges(changes: boolean) {
+    console.log(changes);
```
linusnorton: Probably don't want this in there

projects/media-viewer/src/lib/annotations/comment-set/comment-set.component.ts
velimd: with a service? otherwise how would that pass information outwards?

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -88,8 +89,15 @@ export class CommentComponent implements OnChanges {
     this.editable = true;
   }
 
+  onCommentChange(updatedComment) {
+    const hasUnsavedChanges = this.originalComment.substring(0, this.COMMENT_CHAR_LIMIT) !==
```
velimd: would this not be better to check if the length of the comment character changed is above the character length. The reason for this is i believe if the user added words and was still under the char limit, even tho it will save, this condition will still be true for hasUnsavedChanges but in fact it should save.

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -100,6 +108,7 @@ export class CommentComponent implements OnChanges {
   public onSave() {
     this._comment.content = this.fullComment.substring(0, this.COMMENT_CHAR_LIMIT);
     this.updated.emit(this._comment);
+    this.unsavedChanges.emit(false);
```
velimd: i think this might not be the best place for this. including the one in on cancel, my thinking is, even if you click save, you dont know if the api called passed or failed so the comment will be there and this would emit incorrectly.

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.html
```diff
@@ -23,7 +23,7 @@
               [ngClass]="commentStyle()"
               [(ngModel)]="fullComment"
               [readOnly]="!editable"
-              (ngModelChange)="reRenderComments()"
+              (ngModelChange)="reRenderComments(); onCommentChange($event);"
```
velimd: should this method not be on when the user clicks save? this will be called every time the user is typing.

projects/media-viewer/src/lib/annotations/annotations.module.ts
```diff
@@ -36,7 +36,7 @@ import { CommentSetToggleComponent } from './comment-set/comment-set-toggle/comm
   ],
   providers: [
     AnnotationApiService,
-    AnnotationService
+    AnnotationService,
```
velimd: dont think a comma is needed (not sure why this was a change)

projects/media-viewer/src/lib/annotations/comment-set/comment-set.component.ts
```diff
@@ -196,4 +201,8 @@ export class CommentSetComponent implements OnInit, OnDestroy {
   clearSelection() {
     this.selectAnnotation = { annotationId: '', editable: false };
   }
+
+  allCommentsSaved() {
+    this.commentService.onCommentChange(this.commentComponents.some(comment => comment.editable === true));
```
velimd: i like this logic however, there might be an issue with this (which isnt your change, just need to fix that bug). for example, i believe if youre editing and you click away, it closes the comment and it becomes false. tho it doesnt save, it will still mess with this logic.

projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -58,14 +63,16 @@ export class MediaViewerComponent implements OnChanges, AfterContentInit {
   ngAfterContentInit() {
     this.setToolbarButtons();
     this.toolbarEventsOutput.emit(this.toolbarEvents);
+    this.subscriptions.push(this.commentService.getUnsavedChanges()
```
velimd: if this is the only subscription, might not be best to have it as an array.

PR #207, opened 2019-11-22
making sure comments dont go beyond comments pane by sorting the comm…

pawelkaczmarekuk:
…ents second time in reverse order


projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -135,10 +135,19 @@ export class CommentComponent implements OnChanges {
...
+      'z-index': this.selected ? 100 : 0
     };
   }
 
+  get commentBottomPos(): number {
+    return this.rectTop + this.height;
+  }
+
+  get height() {
+    return this.form.nativeElement.getBoundingClientRect().height / this.zoom;
```
velimd: would the height change even if zoomed? i know top position will be effected but i dont see height changing (unless expanded/collapse)

projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
```diff
@@ -50,9 +54,25 @@ export class CommentSetRenderService {
...
   }
 
   private difference(a: number, b: number): number { return Math.abs(a - b); }
+
+  private makeSureWithinContainer(commentItem: CommentComponent, previousCommentItem: CommentComponent, containerHeight: number) {
+    containerHeight -= 10;
+    if (commentItem.commentBottomPos > containerHeight) {
+      commentItem.rectTop -= commentItem.commentBottomPos - containerHeight;
+    }
+    if (previousCommentItem) {
```
benouaer: I feel like this should be an if-else because the first if is only relevant the first-time when prevComment is null. all subsequent times, prevComment has a value and bottom comment has already been reset, so no need for it to be checked again

projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
```diff
@@ -50,9 +54,25 @@ export class CommentSetRenderService {
...
+      if (commentItem.commentBottomPos > previousCommentItem.rectTop) {
+        commentItem.rectTop -= commentItem.commentBottomPos - previousCommentItem.rectTop;
+      }
+    }
+    if (commentItem.rectTop < 0 ) {
+      commentItem.rectTop = 0;
```
benouaer: I guess we might end up with many comments set to rectTop=0, but there's nothing else we can do if there are too many

pawelkaczmarekuk: correct

projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
pawelkaczmarekuk: I'll make the change

PR #216, opened 2019-12-11
Feature/em 2189 select text native and annotations

velimd:
https://tools.hmcts.net/jira/browse/EM-2189

### Change description ###
To allow text and native annotation to be selected when annotations are created. There is an issue when rotated that the transform style used to rotated the annotation set effect layering/z index, therefore the feature is only allowed when in original position.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
+++ b/projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
@@ -312,7 +312,7 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
   annotationSetClass() {
     return [
       'rotation rot' + this.rotate,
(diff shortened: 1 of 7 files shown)


projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -308,4 +308,11 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
       width: width
     };
   }
+
+  annotationSetClass() {
```
benouaer: you could possibly create a test for this, as codecov is complaining

velimd: :+1:

PR #222, opened 2020-01-03
fixed issue with comments alignment

velimd:
https://tools.hmcts.net/jira/browse/EM-2435

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
+++ b/projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
@@ -34,15 +34,14 @@ export class CommentSetRenderService {
   }
 
   private processSort(a: CommentComponent, b: CommentComponent): number {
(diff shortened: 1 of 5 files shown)


projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
```diff
@@ -6,7 +6,7 @@ export class CommentSetRenderService {
 
   redrawComponents(commentComponents: CommentComponent[], height: number, rotate: number, zoom: number) {
       let prevComment: CommentComponent;
-      this.sortComponents(commentComponents, height, rotate).forEach((comment: CommentComponent) => {
+      this.sortComponents(commentComponents, height / zoom, rotate).forEach((comment: CommentComponent) => {
```
benouaer: can we pass the zoom value and use it in the calculation, because it might be confusing passing it like this

PR #226, opened 2020-01-20
getting rid of padding and transformation when highlighting.

velimd:

https://tools.hmcts.net/jira/browse/EM-2159

### Change description ###

Enabling enhanced textlayer has messed with the creation of highlights due to adding padding and transform dynamically which my fix removes those when creating the highlight.

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
+++ b/projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
@@ -46,6 +46,7 @@ export class PdfViewerComponent implements AfterContentInit, OnChanges, OnDestro
   @Input() height: string;
 
   highlightMode: BehaviorSubject<boolean>;
+  highlighting = false;
   drawMode: BehaviorSubject<boolean>;
 
   loadingDocument = false;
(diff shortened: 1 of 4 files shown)


projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -187,9 +188,14 @@ export class PdfViewerComponent implements AfterContentInit, OnChanges, OnDestro
...
   onMouseUp(mouseEvent: MouseEvent) {
     if (this.toolbarEvents.highlightModeSubject.getValue()) {
-      setTimeout(() => this.viewerEvents.textSelected({
-            page: this.pdfWrapper.getPageNumber(), event: mouseEvent
-      }), 0);
+      this.highlighting = true;
+      setTimeout(() => {
+        this.viewerEvents.textSelected({
+          page: this.pdfWrapper.getPageNumber(),
+          event: mouseEvent
```
benouaer: what about changing the padding value on the target of the event, or would that basically do the same thing?!

velimd: the target event doesnt effect the whole selection, just the last mouse event.

PR #228, opened 2020-01-22
Bug/em 2159 enhanced textlayer highlighting

velimd:
https://tools.hmcts.net/jira/browse/EM-2159

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
+++ b/projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
@@ -129,10 +129,13 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
       if (window.getSelection()) {
         const localElement = (<HTMLElement>highlight.event.target) || (<HTMLElement>highlight.event.srcElement);
 
-        localElement.parentElement.childNodes.forEach(child => {
-          child['style']['padding'] = 0;
-          child['style']['transform'] = child['style']['transform'].replace(/translate[A-Z]\(-?\d*(\.\d+)?(px)?\)/g, '');
-        });
+        if (localElement.parentElement.childNodes) {
+          localElement.parentElement.childNodes.forEach(child => {
+            child['style']['padding'] = 0;
+            const translateCSSRegex = /translate[A-Z]\(-?\d*(\.\d+)?(px)?\)/g;
+            child['style']['transform'] = child['style']['transform'].replace(translateCSSRegex, '');
+          });
+        }
         const selection = window.getSelection();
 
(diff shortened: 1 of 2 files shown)


projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -127,14 +127,24 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
   async createTextHighlight(highlight: Highlight) {
     if (highlight.page === this.page) {
       if (window.getSelection()) {
+        const localElement = (<HTMLElement>highlight.event.target) || (<HTMLElement>highlight.event.srcElement);
```
linusnorton: Maybe: 

```
const localElement: HTMLElement = highlight.event.target || highlight.event.srcElement;
```

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -127,14 +127,24 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
...
+
+        if (localElement.parentElement.childNodes) {
+          localElement.parentElement.childNodes.forEach(child => {
+            child['style']['padding'] = 0;
+            // regex will be targeting the translate style in string
+            // e.g. scaleX(0.969918) translateX(-110.684px) translateY(-105.274px) will become scaleX(0.969918)
+            const translateCSSRegex = /translate[XYZ]\(-?\d*(\.\d+)?(px)?\)/g;
+            child['style']['transform'] = child['style']['transform'].replace(translateCSSRegex, '');
+          });
+        }
```
linusnorton: I would just do:

```
for (const child of localElement.parentElement.childNodes || []) {
  // ...
}
```

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
velimd: this is not possible, as it doesn't covert it into HTMLElement, it just rejects it.

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
velimd: Type 'NodeListOf<ChildNode>' must have a '[Symbol.iterator]()' method that returns an iterator.

linusnorton: :man_facepalming: their array like object doesn't implement Symbol.iterator

PR #230, opened 2020-01-27
Feature/em 1587 grab-n-drag

velimd:
https://tools.hmcts.net/jira/browse/EM-1587

### Change description ###
Having Drag Mode to scroll for the image

```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/directives/drag-scroll.directive.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { DragScrollDirective } from './drag-scroll.directive';
+
+describe('DragScrollDirective', () => {
+  it('should create an instance', () => {
+    const directive = new DragScrollDirective();
```
benouaer: this test doesn't do much. if you want to test the setup, create it with the testbed

projects/media-viewer/src/lib/directives/drag-scroll.directive.ts
```diff
@@ -0,0 +1,49 @@
...
+@Directive({
+  selector: '[dragScroll]'
+})
+export class DragScrollDirective {
+
+  originalScroll: Scroll;
+  private pointerDown = false;
+
+  @Input() enableScroll = true;
+  @Input() scrollAxis = '';
```
benouaer: I would make this explicitly, scrollX, and scrollY, and if it's set it means it's enabled

projects/media-viewer/src/lib/directives/drag-scroll.directive.spec.ts
velimd: i havent done any test yet, first code commit then test commit

projects/media-viewer/src/lib/directives/drag-scroll.directive.ts
```diff
@@ -0,0 +1,49 @@
+import { Directive, ElementRef, HostListener, Input } from '@angular/core';
```
benouaer: I don't think we should have a separate folder for directives, they should just go where they're used. so if this is used with viewers it should just go there. that fits better with angular's domain driven approach.

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.html
```diff
@@ -16,6 +16,11 @@
               data-l10n-id="page_rotate_cw" (click)="rotateCw()">
         <span data-l10n-id="page_rotate_cw_label">Rotate Clockwise</span>
       </button>
+      <button *ngIf="toolbarButtons.showRotate" class="secondaryToolbarButton dragScroll" title="Drag Scroll" tabindex="6"
```
benouaer: why does this use showRotate?

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
```diff
@@ -118,4 +119,13 @@ export class ToolbarEventService {
   public toggleCommentsSummary(value: boolean): void {
     this.showCommentSummary.next(value);
   }
+
+  // Function to toggle dragScroll
+  public toggleDragScroll(): void {
+    if (this.dragScroll.getValue() === false) {
```
benouaer: why not just emit the opposite of the current value

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.html
```diff
@@ -1,9 +1,12 @@
 <mv-comment-set-toggle *ngIf="enableAnnotations"></mv-comment-set-toggle>
-<div id="viewer-wrapper"
+<div dragScroll [scrollAxis]="'horizontal'" [enableScroll]="enableDragScroll"
```
benouaer: do we need two inputs here? would it not work by binding the value of the drag toggle with the scrollAxis (which I think in this case should be dragX), so if it's on dragX would be true otherwise it's false

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.scss
```diff
@@ -60,3 +60,16 @@ img {
   -o-transform-origin: left top;
   transform-origin: left top;
 }
+
+.dragScroll {
```
benouaer: is it possible to add css to a directive?

projects/media-viewer/src/lib/annotations/annotation-set/annotation/rectangle/rectangle.component.html
```diff
@@ -4,7 +4,7 @@
        draggable
        resizable
        [tabindex]="rectangle.id"
-       [ngClass]="{ 'selected': selected }"
+       [ngClass]="{ 'selected': selected, 'dragScroll': enableDragScroll }"
```
benouaer: this is a general comment about the directive, why is it called dragScroll, it doesn't scroll, you grab and move around, so it's more of a grabDrag then dragScroll (I'm not saying call it grabDrag :smile:)

projects/media-viewer/src/lib/directives/drag-scroll.directive.ts
velimd: the folder decision was because, i realised we are starting to have multiple directives so i thought ill create a folder

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.html
velimd: at the time i didnt create a toolbar event for the drag

projects/media-viewer/src/lib/viewers/image-viewer/image-viewer.component.scss
velimd: nope but i think i can just add it to the media viewer and as it doesnt get override it will work, i believe

projects/media-viewer/src/lib/annotations/annotation-set/annotation/rectangle/rectangle.component.html
tomxelliott: grabDrag 🤔

projects/media-viewer/src/lib/toolbar/main-toolbar/sub-toolbar/sub-toolbar.component.html
```diff
@@ -16,6 +16,11 @@
               data-l10n-id="page_rotate_cw" (click)="rotateCw()">
         <span data-l10n-id="page_rotate_cw_label">Rotate Clockwise</span>
       </button>
+      <button *ngIf="toolbarButtons.showGrabNDragButton" [class.toggled]="toolbarEvents.grabNDrag | async" class="secondaryToolbarButton grabNDrag" title="Grab N Drag" tabindex="6"
+              id="grabNDrag"
+              data-l10n-id="page_drag_scroll" (click)="grabNDrag()">
+        <span data-l10n-id="page_drag_scroll_label">Grab N Drag</span>
```
benouaer: page_grab_drag

PR #231, opened 2020-02-04
fixing replace by removing spaces

velimd:
### Change description ###
when you try to replace a string with an empty string with spaces, it doesnt replace the original string. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/media-viewer/src/lib/annotations/annotation-set/annotation-create/text-highlight-create.service.ts
+++ b/projects/media-viewer/src/lib/annotations/annotation-set/annotation-create/text-highlight-create.service.ts
@@ -23,7 +23,10 @@ export class TextHighlightCreateService {
           // regex will be targeting the translate style in string
           // e.g. scaleX(0.969918) translateX(-110.684px) translateY(-105.274px) will become scaleX(0.969918)
           const translateCSSRegex = /translate[XYZ]\(-?\d*(\.\d+)?(px)?\)/g;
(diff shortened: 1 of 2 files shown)


projects/media-viewer/src/lib/annotations/annotation-set/annotation-create/text-highlight-create.service.ts
```diff
@@ -23,7 +23,10 @@ export class TextHighlightCreateService {
           // regex will be targeting the translate style in string
           // e.g. scaleX(0.969918) translateX(-110.684px) translateY(-105.274px) will become scaleX(0.969918)
           const translateCSSRegex = /translate[XYZ]\(-?\d*(\.\d+)?(px)?\)/g;
-          child['style']['transform'] = child['style']['transform'].replace(translateCSSRegex, '');
+          const spaceRegex = /\s+/g;
+          child['style']['transform'] = child['style']['transform']
+            .replace(translateCSSRegex, '')
+            .replace(spaceRegex, '');
```
linusnorton: maybe just `.replace(" ", "")`?

velimd: the regex allows it to detect more than 1 space. i guess the other way is trim() but i know the .replace is not the way to go here

linusnorton: `.replace(" ", "")` will replace all instances of " "

PR #232, opened 2020-02-05
Fix comment panel

velimd:
### Change description ###
cleaning how we inject the comment panel on the media viewer.

```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
```diff
@@ -134,11 +136,20 @@ export class CommentComponent implements OnChanges {
...
     };
   }
 
+  getTotalPreviousPagesHeight(): number {
+    const pageMarginBottom = 10;
+    let totalPreviousPagesHeight = 0;
+    for (let i = 0; i < this.page - 1; i++) {
+      totalPreviousPagesHeight += this.pageHeights[i] + pageMarginBottom;
+    }
+    return totalPreviousPagesHeight;
```
linusnorton: let's go functional!

```
return this.pageHeights.reduce((total, height) => total + height + pageMarginBottom, 0);
```

projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
```diff
@@ -4,22 +4,25 @@ import { CommentComponent } from './comment/comment.component';
...
+    this.pageHeights = pageHeights;
+    let prevComment: CommentComponent;
+    this.sortComponents(commentComponents, rotate, zoom).forEach((comment: CommentComponent) => {
+      this.adjustIfOverlapping(comment, prevComment, zoom);
+      prevComment = comment;
+    });
+    prevComment = null;
+    commentComponents.reverse().forEach((comment: CommentComponent) => {
+      prevComment = this.makeSureWithinContainer(comment, prevComment, containerHeight);
+    });
```
linusnorton: This method could probably be better:

```
redrawComponents(blah...) {
  this.pageHeights = pageHeights; // little bit of vomit here but we're good
  this.sortComponents(commentComponents, rotate, zoom); // sort method mutates 

  for (let i = 1; i < commentComponents.length; i++) {
    this.adjustIfOverlapping(commentComponents[i], commentComponents[i - 1], zoom);
  }
  
  for (let i = commentComponents.length - 2; i >= 0; i++) {
    this.makeSureWithinContainer(commentComponents[i], commentComponents[i + 1], zoom);
  }
```

I'm fairly confident if I knew what `adjustIfOverlapping` and `makeSureWithinContainer` it might be possible to make it better. There's lots of mutable objects here which feels a bit ugly, but it's also how CSS/Dom stuff is.

Main main objections here is `.forEach` - ain't no reason for that. Just use a `for (const derp of derpy) ` it's much easier to read and doesn't involve the overhead of a lambda.

linusnorton: Just noticed that `makeSureWithinContainer` returns the `prevComment` so my example above may not be valid, even so, it seems that the prevComment variable should be avoidable.

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
velimd: just curious, can you stop at a certain index in the array with reduce? if not would it still be better to use reduce but have the function contain an condition?

projects/media-viewer/src/lib/annotations/comment-set/comment-set-render.service.ts
velimd: ill look into that, I believe pawel created it to push the comments up instead of making it scrollable.

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.ts
linusnorton: the reduce function gives three args (total, height, index) so you could use the index, but I'd say it would be better to do:

```
return this.pageHeights
  .slice(0, -1)
  .reduce((total, height) => total + height + pageMarginBottom, 0);
```

PR #234, opened 2020-02-10
set highilght page so that box-highlight is added to only the right page

benouaer:
https://tools.hmcts.net/jira/browse/EM-2597

### Change description ###
fix issue with box-highlights being duplicated on multiple pages

```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/annotations/annotation-set/annotation-create/box-highlight-create.component.spec.ts
```diff
@@ -132,7 +132,7 @@ describe('BoxHighlightCreateComponent', () => {
     component.top = 50;
     component.left = 50;
 
-    component.create();
+    component.create(1);
```
fazanki: not sure if this create is declarative enough maybe createHightlight would make more sense also 1 maybe stick ti into a var called page

benouaer: I'll update it

PR #242, opened 2020-02-24
Feature/em 2607 private icon 

fazanki:
https://tools.hmcts.net/jira/browse/EM-2607


projects/media-viewer/src/lib/annotations/comment-set/comment/comment.component.scss
```diff
@@ -89,3 +92,7 @@ textarea.view-mode {
   white-space: nowrap;
   margin-bottom: 10px;
 }
+
+.aui-comment__meta {
```
velimd: might need to look into why we have some aui-comment styles here and others in the document/media scss.

fazanki: i know it's a bit messy but that will all fall in place i don't know why do we have document.css and why do we need to build the package for the CSS to get picked up. having said that this is all changing with the CSS refactor so will have to come back to this and move it away from here to the external css file

PR #259, opened 2020-04-06
Em 2710 mv e2e tests

kasi-subramaniam:

EM-2710 - Added mv tests - e2e 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/e2e/src/features/comments.panel.feature
+++ b/e2e/src/features/comments.panel.feature
@@ -5,21 +5,17 @@ Feature: CommentsPanelSearch
     Given I am on Media Viewer Page
     Then I expect the page header to be "Media Viewer Demo"
 
-
-#  this works
-  @CommentsPanelSearch
+  @CommentsPanelSearch @ci
   Scenario: Show And Hide the Comments Panel
     When The user clicks on the show comments panel toggle icon
     And  The user clicks to hide the toggle icon
     Then The user clicks on the show comments panel toggle icon
     Then I expect to be able to click on the Filter Search And Comments Tab
     And  I do a invalid search of the Comments Tab
-    #Then The search results are empty
 
-  @CommentsPanelCollateComments
+  @CommentsPanelCollateComments @ci
   Scenario: Collate Comments Overlay Panel
     When The user clicks on the show comments panel toggle icon
     And  The user clicks on the Comments Tab And Then clicks on Collate Summary
     And  There are no comment rows present
-    #Then The user closes the overlay panel
 
--- a/e2e/src/features/comments.panel.feature
+++ b/e2e/src/features/comments.panel.feature
@@ -8,9 +8,7 @@ Feature: CommentsPanelSearch
   @CommentsPanelSearch @ci
   Scenario: Show And Hide the Comments Panel
     When The user clicks on the show comments panel toggle icon
-    And  The user clicks to hide the toggle icon
-    Then The user clicks on the show comments panel toggle icon
-    Then I expect to be able to click on the Filter Search And Comments Tab
+    And  I expect to be able to click on the Filter Search And Comments Tab
     When I Search for Comments
     Then No matching results have been found
 
--- a/e2e/src/features/comments.panel.feature
+++ b/e2e/src/features/comments.panel.feature
@@ -1,0 +1,25 @@
+@MediaViewer @Search
+Feature: CommentsPanelSearch
+
+  Background:
+    Given I am on Media Viewer Page
+    Then I expect the page header to be "Media Viewer Demo"
+
+
+#  this works
+  @CommentsPanelSearch
+  Scenario: Show And Hide the Comments Panel
+    When The user clicks on the show comments panel toggle icon
+    And  The user clicks to hide the toggle icon
+    Then The user clicks on the show comments panel toggle icon
+    Then I expect to be able to click on the Filter Search And Comments Tab
+    And  I do a invalid search of the Comments Tab
+    #Then The search results are empty
+
+  @CommentsPanelCollateComments
+  Scenario: Collate Comments Overlay Panel
+    When The user clicks on the show comments panel toggle icon
+    And  The user clicks on the Comments Tab And Then clicks on Collate Summary
+    And  There are no comment rows present
+    #Then The user closes the overlay panel
+
--- a/e2e/src/features/comments.panel.feature
+++ b/e2e/src/features/comments.panel.feature
@@ -11,11 +11,14 @@ Feature: CommentsPanelSearch
     And  The user clicks to hide the toggle icon
     Then The user clicks on the show comments panel toggle icon
     Then I expect to be able to click on the Filter Search And Comments Tab
-    And  I do a invalid search of the Comments Tab
+    When I Search for Comments
+    Then No matching results have been found
+
 
   @CommentsPanelCollateComments @ci
   Scenario: Collate Comments Overlay Panel
-    When The user clicks on the show comments panel toggle icon
-    And  The user clicks on the Comments Tab And Then clicks on Collate Summary
-    And  There are no comment rows present
+    When  The user clicks on the show comments panel toggle icon
+    And   The user clicks on the Comments Tab
+    And   The user clicks on Collate Summary
+    Then  There are no comment rows present
 
--- a/e2e/src/pages/commentspanel.po.ts
+++ b/e2e/src/pages/commentspanel.po.ts
@@ -1,0 +1,103 @@
+import {browser, by, element, ElementArrayFinder, ElementFinder,} from 'protractor';
+import {AppPage} from './app.po';
+import {By} from '@angular/platform-browser';
+import {expect} from 'chai';
+
+
+export class CommentsPanelPage extends AppPage {
+  //commentsToggleButton : ElementFinder = element(by.id('toggle-comments-panel'));
+
+  commentsToggleButton : ElementFinder      = element(by.xpath("//button[@id=\'toggle-comments-panel\' and @title=\'Toggle Comments Panel\']"));
+  commentsTabButton     : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[1]/a")));
+  filterTabButton       : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[2]/a")));
+  searchTabButton       : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[3]/a")));
+  collateCommentsButton : ElementFinder     = element(by.xpath("//button[@id=\'commentSummary\']"));
+  commentsNotAvailable  : ElementFinder     = element(by.xpath("//div[@id=\'comment-container\'/h3"));
+  courtBundleName       : ElementFinder     = element(by.xpath("//div[@id=\'comment-container\'/h2"));
+  overlayCloseButton    : ElementFinder     = element(by.buttonText("Close"));
+  searchInputField      : ElementFinder      = element(by.id('search-comments-input'));
+  searchButton          : ElementFinder     = element(by.xpath('//button[@class=\'govuk-button comment-search\']'));
+  searchResult          : ElementFinder     = element(by.css('#comment-search'));
+
+  async clickShowCommentsToggleIcon(){
+     await this.commentsToggleButton.click();
+
+  }
+
+  async clickOnCommentsTab(){
+    await browser.sleep(1000);
+    this.commentsTabButton.getText().then( function ( name ) {
+      expect(name).equal('Comments');
+
+    })
+    await this.commentsTabButton.click();
+  }
+
+  async clickOnFilterTab(){
+    this.filterTabButton.getText().then( function ( name ) {
+      expect(name).equal('Filter');
+
+    })
+    await this.filterTabButton.click();
+  }
+
+  async clickOnSearchTab(){
+    await browser.sleep(5000);
+    await this.searchTabButton.click();
+    await browser.sleep(5000);
+ }
+
+  async performSearch(){
+    let searchText = "asdfasdfadsfasf";
+
+    await browser.sleep(1000);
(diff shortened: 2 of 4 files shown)


e2e/src/features/comments.panel.feature
```diff
@@ -0,0 +1,21 @@
...
+    When The user clicks on the show comments panel toggle icon
+    And  The user clicks to hide the toggle icon
+    Then The user clicks on the show comments panel toggle icon
+    Then I expect to be able to click on the Filter Search And Comments Tab
+    And  I do a invalid search of the Comments Tab
+
+  @CommentsPanelCollateComments @ci
+  Scenario: Collate Comments Overlay Panel
+    When The user clicks on the show comments panel toggle icon
+    And  The user clicks on the Comments Tab And Then clicks on Collate Summary
```
benouaer, line 19: I think clicking on the comments tab and then on collate summary should be two steps

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
+import {browser, by, element, ElementArrayFinder, ElementFinder,} from 'protractor';
+import {AppPage} from './app.po';
+import {By} from '@angular/platform-browser';
+import {expect} from 'chai';
+
+
+export class CommentsPanelPage extends AppPage {
+  //commentsToggleButton : ElementFinder = element(by.id('toggle-comments-panel'));
+
+  commentsToggleButton : ElementFinder      = element(by.xpath("//button[@id=\'toggle-comments-panel\' and @title=\'Toggle Comments Panel\']"));
```
benouaer, line 10: why do we need to check for the title here?

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+  commentsTabButton     : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[1]/a")));
+  filterTabButton       : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[2]/a")));
+  searchTabButton       : ElementFinder     = element(by.xpath(("//div[@class=\'govuk-tabs commentSummaryHeader\']/ul/li[3]/a")));
```
benouaer, line 13: I wouldn't rely on the order, it's better if there are Ids or css classes
also, it seems to me like searching by xpath like this would be very sensitive to the smallest change in layout. why not search by specific id or class

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+  collateCommentsButton : ElementFinder     = element(by.xpath("//button[@id=\'commentSummary\']"));
+  commentsNotAvailable  : ElementFinder     = element(by.xpath("//div[@id=\'comment-container\'/h3"));
+  courtBundleName       : ElementFinder     = element(by.xpath("//div[@id=\'comment-container\'/h2"));
+  overlayCloseButton    : ElementFinder     = element(by.buttonText("Close"));
+  searchInputField      : ElementFinder      = element(by.id('search-comments-input'));
+  searchButton          : ElementFinder     = element(by.xpath('//button[@class=\'govuk-button comment-search\']'));
+  searchResult          : ElementFinder     = element(by.css('#comment-search'));
+
+  async clickShowCommentsToggleIcon(){
+     await this.commentsToggleButton.click();
```
benouaer, line 23: I think naming this function clickShowCommentsToggleIcon is misleading, because it might show or hide depending on the current state

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  }
+
+  async clickOnCommentsTab(){
+    await browser.sleep(1000);
+    this.commentsTabButton.getText().then( function ( name ) {
```
benouaer, line 29: it's better if we use async await rather than then

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+      expect(name).equal('Comments');
+
+    })
+    await this.commentsTabButton.click();
+  }
+
+  async clickOnFilterTab(){
+    this.filterTabButton.getText().then( function ( name ) {
```
benouaer, line 37: replace with async await if possible

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+      expect(name).equal('Filter');
+
+    })
+    await this.filterTabButton.click();
+  }
+
+  async clickOnSearchTab(){
+    await browser.sleep(5000);
```
benouaer, line 45: this really slows down the test if we're waiting 5secs each time. it would better to wait for actions if possible

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+    await this.collateCommentsButton.click();
+    await browser.sleep(3000);
+  }
+
+  async hideCommentsToggle(){
+    await this.commentsToggleButton.click();
+ }
+
+ async assertToggleButtonText()  {
+   await this.commentsToggleButton.getText().then( function(toggleIconText) {
```
benouaer, line 74: no need to use then if we're using async await

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+     expect(toggleIconText).equal('Hide comments');
+   })
+ }
+
+  async assertSearchResultText(searchResultStringOnPanel :String)  {
+    await this.searchResult.getText().then( function(text) {
```
benouaer, line 80: again, no need to mix the use of async/await with then

e2e/src/pages/commentspanel.po.ts
```diff
@@ -0,0 +1,103 @@
...
+      expect(text.trim()).equal(searchResultStringOnPanel);
+    })
+  }
+
+  async assertNoCommentRowsPresent() {
+    this.commentsNotAvailable.getText().then( function ( name ) {
```
benouaer, line 86: we can use async/await here as well

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.clickShowCommentsToggleIcon();
+  await genericMethods.sleep(3000);
+
+  // await commentsPanelPage.clickOnSearchTab();
+
+});
+
+//I expect to see the comments filter and search tabs
+When(/^I expect to be able to click on the Filter Search And Comments Tab$/, async function () {
```
benouaer, line 470: when should be used with actions, like the user clicking on something. expectations should be done in then

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+
+When(/^I do a invalid search of the Comments Tab$/, async function () {
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.performSearch();
+
+});
+
+
+
+When(/^The user enters a random search string$/, async function () {
```
benouaer, line 488: I don't see where the user enters the random search string here

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.clickShowCommentsToggleIcon();
+
+  console.log('media viewer comments tab is  found');
+  const viewableDoc = await browser.takeScreenshot();
+  this.attach(viewableDoc, 'image/png');
+
+});
+
+When(/^The user clicks on the search button of the comments panel$/, async function () {
```
benouaer, line 498: again, I don't see where the user is clicking on the search button in this function

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.clickShowCommentsToggleIcon();
+
+  console.log('media viewer comments tab is  found');
+  const viewableDoc = await browser.takeScreenshot();
+  this.attach(viewableDoc, 'image/png');
+
+});
+
+When(/^The search results are empty$/, async function () {
```
benouaer, line 508: this should be either a Given for when the user starts with no search results, or a then if it's the result of a search that the user has just done

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  //TODO Logic
+  let noMatchText ="No matches have been found".trim();
+  await commentsPanelPage.assertSearchResultText(noMatchText);
+  const viewableDoc = await browser.takeScreenshot();
+
+  this.attach(viewableDoc, 'image/png');
+
+});
+
+Then(/^The user should see that comments panel is visible$/, async function () {
```
benouaer, line 519: the description of the function doesn't seem to match the content

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  await genericMethods.sleep(1000);
+  await commentsPanelPage.hideCommentsToggle();
+  const viewableDoc = await browser.takeScreenshot();
+  this.attach(viewableDoc, 'image/png');
+});
+
+When(/^The user clicks on the Comments Tab And Then clicks on Collate Summary$/, async function () {
```
benouaer, line 526: I think this should be split into two

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.clickOnCommentsTab();
+  await genericMethods.sleep(3000);
+
+  await commentsPanelPage.clickOnCollateCommentsButton();
+
+});
+
+When(/^There are no comment rows present$/, async function () {
```
benouaer, line 535: again, description doesn't seem to match the content

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -452,3 +456,98 @@ Then('I expect custom toolbar button should be enabled', async () => {
...
+});
+
+When(/^The user closes the overlay panel$/, async function () {
+  await genericMethods.sleep(2000);
+  await commentsPanelPage.closeOverlayPanel()
+
+});
+
+
+Then(/^The user clicks to hide the toggle icon$/, async function () {
```
benouaer, line 550: user actions should be in when clauses

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: removed the check for title .

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: Unable to do this at the moment as the only way I can assert ordering of the tabs is through the array elements.

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: Renamed to clickCommentsToggleIcon

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: calls to  .getText().then replaced with async await.

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: Search String is now passed in as a parameter to the performSearch() method.

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: Removed redundant method.

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: refactored

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: refactored

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: refactored

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: refactored

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: this has been done now.

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: refactored.

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: removed the  sleep calls in several places.

e2e/src/pages/commentspanel.po.ts
kasi-subramaniam: removed this now

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: Removed this function.

e2e/src/step_definitions/media-viewer.ts
kasi-subramaniam: this has now been resolved.

PR #263, opened 2020-04-15
Feature/em 2613 filter comments 2

fazanki:

### Change description ###

```
[ ] Yes
[ ] No
```


projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-search/commet-search.component.spec.ts
```diff
@@ -91,7 +91,8 @@ describe('CommentSearch', () => {
 
 @Component({
   selector: `host-component`,
-  template: `<mv-comment-search [annotationSet]="annotationSet"></mv-comment-search>`
+  template: `
+      <mv-comment-search [annotations]="annotationSet"></mv-comment-search>`
 })
 class TestHostComponent {
   annotationSet = { annotations: [] };
```
benouaer, line 98: I guess we should pass annotations here now

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-set-header.component.ts
```diff
@@ -1,27 +1,41 @@
...
+     this.isFiltered = !formData.length;
+     this.tabs = [...this.tabs].map((tab) => {
+       return !this.isFiltered && tab.label === 'filter' ? {...tab, isFiltered: true} : {...tab, isFiltered: false};
+     });
+    });
+
+    this.store.pipe(select(fromAnnoSelector.getFilteredAnnotations)).subscribe(filteredAnno => {
+      this.navigationList = filteredAnno.length ?
+        filteredAnno.filter(annotation => annotation.comments && annotation.comments.length > 0) : [];
+
```
benouaer, line 37: is it not worth having a selector for this?

projects/media-viewer/src/lib/annotations/models/tags.model.ts
```diff
@@ -1,4 +1,4 @@
-export interface TagItemModel {
+export interface TagsModel {
```
benouaer, line 1: why not just call it Tags

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/commet-set-header.component.spec.ts
```diff
@@ -1,27 +1,34 @@
...
 import {RouterTestingModule} from '@angular/router/testing';
+import {StoreModule} from '@ngrx/store';
+import {reducers} from '../../../store/reducers';
 
 describe('CommentSetHeader', () => {
   let component: CommentSetHeaderComponent;
   let fixture: ComponentFixture<CommentSetHeaderComponent>;
   let nativeElement;
-  let toolbarEventService: ToolbarEventService
+  // let toolbarEventService: ToolbarEventService
```
benouaer, line 13: I guess this can be removed

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/commet-set-header.component.spec.ts
```diff
@@ -1,27 +1,34 @@
...
+
+      ],
       schemas: [CUSTOM_ELEMENTS_SCHEMA]
     }).compileComponents();
 
     fixture = TestBed.createComponent(CommentSetHeaderComponent);
     nativeElement = fixture.debugElement.nativeElement;
     component = fixture.componentInstance;
-    toolbarEventService = new ToolbarEventService();
+    // toolbarEventService = new ToolbarEventService({});
```
benouaer, line 31: guess this can be removed

projects/media-viewer/src/lib/store/reducers/annotatons.reducer.ts
```diff
@@ -92,7 +92,7 @@ export function reducer (
           id: uuid()
         };
       const annotationEntities = StoreUtils.generateAnnotationEntities(annotationSet.annotations);
-      const annotationPageEntities = StoreUtils.generatePageEntities(annotationSet.annotations);
+      const annotationPageEntities = StoreUtils.groupByKeyEntities(annotationSet.annotations, 'page');
```
benouaer, line 95: like this

projects/media-viewer/src/lib/store/reducers/tags.reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+import * as fromTags from '../actions/tags.actions';
+import {StoreUtils} from '../store-utils';
+import {Annotation} from '../../annotations/annotation-set/annotation-view/annotation.model';
+
+export interface TagsState {
+  tagNameEnt: {[id: string]: string[]};
+  filteredComments: {[id: string]: string[]};
+  filteredPageEntities: {[id: string]: Annotation[]};
+  annotations: Annotation[];
+  formFilterState: {[id: string]: boolean};
```
benouaer, line 11: I assume this should be fromFilterState

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
```diff
@@ -1,5 +1,8 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable, Subject } from 'rxjs';
+import {Store} from '@ngrx/store';
+import * as fromStore from '../store/reducers';
+import * as fromTagActions from '../store/actions/tags.actions';
```
benouaer, line 5: these store imports don't look like they're being used here

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
```diff
@@ -34,6 +37,8 @@ export class ToolbarEventService {
   public readonly changePageByDeltaSubject = new Subject<number>();
   public readonly showCommentSummary = new BehaviorSubject<boolean>(false);
   public readonly grabNDrag = new BehaviorSubject<boolean>(false);
+  constructor() {
+  }
```
benouaer, line 41: this doesn't look like it's needed

projects/media-viewer/src/lib/store/store-utils.ts
```diff
@@ -26,6 +27,52 @@ export class StoreUtils {
       }, {});
   }
 
+  static genTagNameEntities(annotations) {
+    const filterAnnoWithNoCommentsTags = annotations.filter(a => (a.comments.length && a.tags.length));
```
benouaer, line 31: this filter gives you annotations that have comments and tags, so the name is misleading

projects/media-viewer/src/lib/store/store-utils.ts
```diff
@@ -1,19 +1,20 @@
 import {Annotation} from '../annotations/annotation-set/annotation-view/annotation.model';
+
 // @dynamic
 export class StoreUtils {
 
-  static generatePageEntities(annotations): {[id: string]: Annotation[]} {
+  static groupByKeyEntities(annotations, key): {[id: string]: Annotation[]} {
```
benouaer, line 6: I guess this is no longer annotations and Annotation[], because it's also used for tag entities

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-search/commet-search.component.spec.ts
fazanki: its a spec file

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-set-header.component.ts
fazanki: yeap - good spot

projects/media-viewer/src/lib/annotations/models/tags.model.ts
fazanki: dunno maybe because Tags dos not mean is a Model - ti can be an array

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/commet-set-header.component.spec.ts
fazanki: done

projects/media-viewer/src/lib/store/reducers/tags.reducer.ts
fazanki: it is form from FormBuilder

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
fazanki: yeap

projects/media-viewer/src/lib/toolbar/toolbar-event.service.ts
fazanki: ok

projects/media-viewer/src/lib/store/store-utils.ts
fazanki: correct it should be any[]

projects/media-viewer/src/lib/store/store-utils.ts
fazanki: why don't u suggest something then

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/commet-set-header.component.spec.ts
fazanki: done

projects/media-viewer/src/lib/store/store-utils.ts
```diff
@@ -26,6 +27,52 @@ export class StoreUtils {
...
+     return Object.keys(groupedByName).reduce(
+      (tagNameEnt, key) => {
+        const annot = annos.map(anno => {
+          const tags = anno.tags.map(tag => {
+            if (tag.name === key) {
+              return anno.id;
+            }
+          });
+          return tags.filter(a => a !== undefined);
+        });
```
benouaer, line 54: a bit confused by the logic here, but isn't it possible to do it like this: 
        annos.filter(anno => anno.tags.find(tag => tag.name === name))
          .map(anno => anno.id)

projects/media-viewer/src/lib/store/store-utils.ts
```diff
@@ -26,6 +27,52 @@ export class StoreUtils {
...
+            return {
+              ...obj,
+              [a]: a
+            };
+          } else {
+            return {
+              ...obj
+            };
+          }
+        }, {})
```
benouaer, line 67: so is readyAnno an object where the key is the annoId and the value is the annoId

fazanki: i think if u think this can be done better we need to pair up... is much more complex then that... but there might be a better solution... i am not very happy but it works and it gives me what I want - yeah is the brain/logic  of the whole filtering

benouaer: this seems to work:
return Object.keys(groupedByName).reduce(
      (tagNameEnt, key) => {
        const readyAnno = annos.filter(anno => anno.tags.find(tag => tag.name === key))
          .map(anno => anno.id)
          .reduce((obj: {[id: string]: string}, anno: string) => ({ ...obj, [anno]: anno }), {});
        return {
          ...tagNameEnt,
          [key]: readyAnno
        }
      }, {});

PR #268, opened 2020-04-24
Bug/em 2854 improve unsupported doc handling

jackmunkonge:
https://tools.hmcts.net/jira/browse/EM-2854

### Change description ###
Added new error messages for PDF and Image document load errors (error message with download link)

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/media-viewer/src/lib/media-viewer.component.html
+++ b/projects/media-viewer/src/lib/media-viewer.component.html
@@ -31,7 +31,7 @@
                      [downloadFileName]="downloadFileName"
                      [height]="height">
     </mv-image-viewer>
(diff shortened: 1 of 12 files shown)


projects/media-viewer/src/lib/media-viewer.component.html
```diff
@@ -31,8 +31,9 @@
                      [downloadFileName]="downloadFileName"
                      [height]="height">
     </mv-image-viewer>
-    <mv-unsupported-viewer *ngIf="contentTypeUnsupported()"
+    <mv-unsupported-viewer *ngIf="contentTypeUnsupported() || typeException"
```
benouaer, line 34: I think we found that we don't need this part: || typeException

PR #276, opened 2020-05-14
Em 2919 in court presentation

meganaslattery:

https://tools.hmcts.net/jira/browse/EM-2919

### Change description ###

ICP POC using Socketio

```
[ ] Yes
[x] No
```


projects/media-viewer/src/lib/providers/socket.service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+@Injectable()
+export class SocketService {
+
+  private socket: SocketIOClient.Socket;
+  connected$ = new BehaviorSubject<boolean>(false);
+  socketUrl = 'http://localhost:8080';
+
+  constructor() {
+    // @ts-ignore
```
benouaer, line 13: what's the ts-ignore for?

projects/media-viewer/src/lib/icp/icp-session.service.ts
```diff
@@ -0,0 +1,19 @@
...
+@Injectable()
+export class IcpSessionService {
+
+  private ICP_SESSION_API = '/icp/sessions';
+
+  constructor(private readonly httpClient: HttpClient) {}
+
+  public loadSession(caseId: string): Observable<any> {
+    return this.httpClient
+      .get(`${this.ICP_SESSION_API}/${caseId}`,
```
benouaer, line 15: it would be useful to include the return type here with the get 
httpClient.get<>

projects/media-viewer/src/lib/icp/icp-session.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
```
benouaer, line 1: maybe this file could be renamed to include api, as that seems to be consistent with the rest of the app

projects/media-viewer/src/lib/icp/icp.directive.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (this.isPresenter) {
+      this.stopPresenting();
+    }
+    this.socketService.leaveSession();
+    this.store.dispatch(new fromIcpActions.LeaveIcpSocketSession());
+    this.unsubscribeSession();
+  }
+
+  stopPresenting() {
+    const presenter: IcpParticipant = {username: '', id: ''};
```
benouaer, line 84: can we set the whole thing to undefined

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+  client: null
+};
+
+export function icpReducer (state = initialIcpSessionState,
+                                  action: fromIcpActions.IcpActions): IcpState {
+
+  switch (action.type) {
+
+    case fromIcpActions.ICP_SOCKET_SESSION_JOINED: {
+      const session: IcpSession = action.payload.session;
```
benouaer, line 16: might be worth using the deconstruction operation here {...}

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+      return {
+        ...state,
+        session,
+        client: participantInfo.client,
+        presenter: participantInfo.presenter
+      };
+    }
+
+    case fromIcpActions.ICP_PRESENTER_UPDATED: {
+      const presenter = action.payload;
```
benouaer, line 27: maybe use the deconstruction operator here as well

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.html
```diff
@@ -34,13 +34,13 @@
 </div>
 <div id="annotationButtons">
   <div class="splitToolbarButton hiddenIfLargerThan1000px" >
-    <button *ngIf="toolbarButtons.showHighlightButton"
+    <button *ngIf="toolbarButtons.showHighlightButton && !icpSimplifiedToolbar"
```
benouaer, line 37: you could have the ngIf on the outer div, so you only do the check once

projects/media-viewer/src/lib/toolbar/left-pane/left-pane.component.ts
```diff
@@ -8,8 +8,10 @@ import { Subscription } from 'rxjs';
   templateUrl: './left-pane.component.html',
 })
 export class ToolbarLeftPaneComponent implements OnInit, OnDestroy {
+
   public pageNumber = 1;
   private subscriptions: Subscription[] = [];
+  icpSimplifiedToolbar = false;
```
benouaer, line 14: icpEnabled?

projects/media-viewer/src/lib/toolbar/right-pane/right-pane.component.html
```diff
@@ -9,28 +15,31 @@
           data-l10n-id="open_file">
     <span data-l10n-id="open_file_label">Open</span>
   </button>
-  <button *ngIf="enableRedactions"
+  <div class="toolbarButtonSpacer"></div>
+  <button *ngIf="enableRedaction && !icpSimplifiedToolbar"
```
benouaer, line 19: these two checks seem to be doing the same thing, so I think we should be using one of them:
- !(toolbarEvents.icp.enabled | async)
- !icpSimplifiedToolbar

projects/media-viewer/src/lib/toolbar/right-pane/right-pane.component.html
```diff
@@ -9,28 +15,31 @@
...
           id="redact" class="toolbarButton redact hiddenMediumView" title="Redact" tabindex="0"
           data-l10n-id="redact" (click)="toggleRedactionMode()">
     <span data-l10n-id="redact_label">Redact</span>
   </button>
-  <button *ngIf="toolbarButtons.showDownload"
+  <div class="toolbarButtonSpacer"></div>
```
benouaer, line 24: you could include all of these in one ng-container, and hide them all in one go

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.html
```diff
@@ -46,5 +46,13 @@
       <span data-l10n-id="download_label">Download</span>
       </button>
     <div class="horizontalToolbarSeparator"></div>
+    <div  *ngIf="!(icpSession$ | async)">
```
benouaer, line 49: we have the icpEnabled boolean which we can use here

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.ts
```diff
@@ -1,18 +1,29 @@
...
 
   constructor(
     public readonly toolbarButtons: ToolbarButtonVisibilityService,
-    public readonly toolbarEvents: ToolbarEventService
+    public readonly toolbarEvents: ToolbarEventService,
+    private store: Store<IcpState>
   ) {}
 
+  ngOnInit() {
+    this.icpSession$ = this.store.pipe(select(fromSelectors.getIcpSession));
```
benouaer, line 24: we can just check the icpEnabled boolean from toolbar events. that's what we're doing with the other launch button

src/polyfills.ts
```diff
@@ -78,3 +78,7 @@ import 'zone.js/dist/zone';  // Included with Angular CLI.
 /***************************************************************************************************
  * APPLICATION IMPORTS
  */
+
+
+// Add global to window, assigning the value of window itself.
+(window as any).global = window;
```
benouaer, line 84: is this required?

projects/media-viewer/src/lib/icp/icp.directive.ts
meganaslattery: The same updatePresenter event is emitted for stopping presenter or changing presenter.. so the backend expects an IcpParticipant object.

src/polyfills.ts
meganaslattery: Yes, there is a bug with socketio and angular 6+ which this fixes https://medium.com/iamdeepinder/creating-a-real-time-app-with-angular-8-and-socket-io-with-nodejs-af63bd59a47f

projects/media-viewer/src/lib/icp/icp-follower.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!this.$subscription) {
+      this.$subscription = this.socketService.screenUpdated()
+        .subscribe(screen => this.followScreenUpdate(screen));
+    }
+  }
+
+  unsubscribe() {
+    if (this.$subscription) {
+      this.$subscription.unsubscribe();
+      this.$subscription = undefined;
```
fazanki, line 42: is this needed?

projects/media-viewer/src/lib/icp/icp.directive.ts
benouaer: ok, in that case maybe set the username and id as undefined instead of empty string

projects/media-viewer/src/lib/store/reducers/document.reducer.ts
```diff
@@ -79,10 +81,19 @@ export function docReducer (state = initialDocumentState,
         hasDifferentPageSize
       };
     }
+
+    case fromActions.POSITION_UPDATED: {
+      const pdfPosition = action.payload;
+      return {
+        ...state,
+        pdfPosition: pdfPosition
```
fazanki, line 89: u can use just pdfPossition without the :

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+      const presenter = {...action.payload};
+      return {
+        ...state, presenter
+      };
+    }
+
+    case fromIcpActions.LEAVE_ICP_SOCKET_SESSION: {
+      return {
+        ...state,
+        session: null,
```
fazanki, line 36: use ...state,
initialcpSessionState

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+  client: null
+};
+
+export function icpReducer (state = initialIcpSessionState,
+                                  action: fromIcpActions.IcpActions): IcpState {
+
+  switch (action.type) {
+
+    case fromIcpActions.ICP_SOCKET_SESSION_JOINED: {
+      const session: IcpSession = {...action.payload.session};
```
fazanki, line 16: if you need to do this {...} then is better to decouple it in effect so that mutable objects are not sent in action

src/polyfills.ts
benouaer: ok, we would have to include this in the ReadMe, and highlight it in the integration instructions for ICP

projects/media-viewer/src/lib/icp/icp.directive.ts
```diff
@@ -0,0 +1,97 @@
...
+  @Input() caseId: string;
+
+  client: IcpParticipant;
+  presenter: IcpParticipant;
+  isPresenter: boolean;
+
+  subscription: Subscription;
+  sessionSubscription: Subscription;
+
+  constructor(private readonly toolbarEvents: ToolbarEventService,
```
fazanki, line 29: https://angular.io/guide/attribute-directives
this approach is going against the angular guidelines. 
directives are meant to use only for manipulating elements - but i like your thinking

projects/media-viewer/src/lib/icp/icp.directive.ts
```diff
@@ -0,0 +1,97 @@
...
+              private readonly viewerEvents: ViewerEventService,
+              private readonly socketService: IcpUpdateService,
+              private readonly presenterSubscriptions: IcpPresenterService,
+              private readonly followerSubscriptions: IcpFollowerService,
+              private store: Store<IcpState>) {}
+
+  ngOnInit() {
+    this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
```
fazanki, line 37: icp.sessionLaunch is the step that can be removed and just dispatch action where u trigger .next()
if we have ngrx services should be really just for HTTP.

projects/media-viewer/src/lib/toolbar/icp-event.service.ts
```diff
@@ -0,0 +1,34 @@
+import { BehaviorSubject, Subject } from 'rxjs';
```
fazanki, line 1: u have too many steps/services/bridges... try to simplify it

projects/media-viewer/src/lib/icp/icp-follower.service.ts
benouaer: yes, we need to stop listening for position updates if we're no longer followers

projects/media-viewer/src/lib/icp/icp.directive.ts
benouaer: not sure if I understand correctly, but we can't dispatch the action where we do trigger.next() because that happens in a toolbar, and toolbars are optional, so we can't put functionality in there. that's why trigger an event from the toolbar, and then handle it here, in the same way that someone using their own toolbar might do

projects/media-viewer/src/lib/icp/icp.directive.ts
benouaer: we can argue that we're manipulating the component, which is the pdf-viewer

PR #288, opened 2020-06-16
EM-2830 In-document Bookmark Icon

jasonpaige:
### Change description ###
Adds a bookmark icon in the pdf viewer which aligns to where the bookmark was created in the document.

```
[ ] Yes
[X] No
```


projects/media-viewer/src/lib/bookmark/bookmark.module.ts
```diff
@@ -0,0 +1,26 @@
...
+  entryComponents: [
+  ],
+  providers: [
+    BookmarkApiService
+  ],
+  exports: [
+    BookmarkComponent
+  ]
+})
+export class BookmarkModule { }
```
fazanki, line 26: do u really need a module? don't think is a module

projects/media-viewer/src/lib/store/reducers/bookmarks.reducer.ts
```diff
@@ -19,47 +20,66 @@ export function bookmarksReducer (state = initialBookmarksState,
...
 
     case fromBookmarks.CREATE_BOOKMARK_SUCCESS: {
       const bookmark: Bookmark = action.payload;
       const bookmarkEntities = {
         ...state.bookmarkEntities,
         [bookmark.id]: bookmark,
-      }
+      };
+      console.log('Create Bookmark');
+      console.log('=============');
```
fazanki, line 47: do we need console logs

projects/media-viewer/src/lib/store/selectors/bookmarks.selectors.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        div: {},
+        scale: 1,
+        rotation: 0,
+        id: '1'
+      }]));
+
+      let result;
+      store.pipe(select(fromSelectors.getBookmarksPerPage)).subscribe(value => {
+        console.log('=================');
+        console.log(value);
```
fazanki, line 34: don't think we need console logs

projects/media-viewer/src/lib/store/selectors/redaction.selectors.ts
```diff
@@ -44,6 +44,7 @@ export const getRedactionsPerPage = createSelector(
   fromDocument.getPages,
   getRedactionPages,
   (pages, pageEnt) => {
+    console.log(pageEnt);
```
fazanki, line 47: console log

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.html
```diff
@@ -47,6 +47,10 @@
                     [height]="pdfViewer.offsetHeight"
                     [pageHeights]="pageHeights">
     </mv-comment-set>
+    <mv-bookmark #bookmarkComponent
+                 [zoom]="zoom"
```
fazanki, line 51: #bookmarkComponent is this used anywhere

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@Component({
+  selector: 'mv-bookmark',
+  templateUrl: './bookmark.component.html'
+})
+export class BookmarkComponent implements OnInit, OnDestroy {
+
+  @Input() zoom: number;
+  @Input() rotate: number;
+  bookmarksPerPage$: Observable<any>; // todo add type
```
fazanki, line 16: you could add the type

projects/media-viewer/src/lib/bookmark/bookmark.module.ts
jasonpaige: Honestly, I only made it a module as I was basing my work from what had been done with redactions and that had a todo for making it into a module...https://github.com/hmcts/media-viewer/blob/master/projects/media-viewer/src/lib/media-viewer.module.ts#L64
Happy to conform to how redactions are used currently if you prefer?

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
jasonpaige: Could you point me at any examples for this please?

projects/media-viewer/src/lib/bookmark/bookmark.module.ts
fazanki: yeah, I put that in there.... I think we should have a conversation about how to structure the app - atm we have no rules. leave it as it is

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
fazanki: all you need to do is create an interface based on what that observable streams  and put it instead of <any> ping me directly if you need help to find out what it streams

projects/media-viewer/src/lib/annotations/annotation-set/annotation-set.component.ts
```diff
@@ -81,19 +82,22 @@ export class AnnotationSetComponent implements OnInit, OnDestroy {
   }
 
   createBookmark(rectangle: Rectangle) {
-    this.store.pipe(select(fromBookmarks.getBookmarkInfo), take(1))
-      .subscribe((bookmarkInfo) => {
-        const selection = window.getSelection().toString();
+    const selection = window.getSelection().toString();
+    this.store.pipe(select(fromAnnotations.getDocumentIdSetId))
```
benouaer, line 86: you can get the documentId from a document selector rather than annotations, so as not to tie this to annotations

projects/media-viewer/src/lib/bookmark/bookmark.module.ts
```diff
@@ -0,0 +1,26 @@
+import { NgModule } from '@angular/core';
```
benouaer, line 1: not sure we need a separate module for bookmarks, especially not just for this component

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
```diff
@@ -0,0 +1,32 @@
...
+import { select, Store } from '@ngrx/store';
+import * as fromStore from '../../store/reducers/reducers';
+import * as fromSelectors from '../../store/selectors/bookmarks.selectors';
+import {Observable, Subscription} from 'rxjs';
+
+@Component({
+  selector: 'mv-bookmark',
+  templateUrl: './bookmark.component.html'
+})
+export class BookmarkComponent implements OnInit, OnDestroy {
```
benouaer, line 12: could we rename this component to something like bookmark-icon or similar, to differentiate it from the bookmarks component that represents the actual bookmarks

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  private $subscription: Subscription;
+
+  constructor(private store: Store<fromStore.State>) {}
+
+  ngOnInit(): void {
+    this.bookmarksPerPage$ = this.store.pipe(select(fromSelectors.getBookmarksPerPage));
+  }
+
+  ngOnDestroy(): void {
+    this.$subscription.unsubscribe();
```
benouaer, line 29: no need for this, as you're not subscribing to any observables

projects/media-viewer/src/lib/bookmark/components/bookmark.component.ts
```diff
@@ -0,0 +1,32 @@
...
+})
+export class BookmarkComponent implements OnInit, OnDestroy {
+
+  @Input() zoom: number;
+  @Input() rotate: number;
+  bookmarksPerPage$: Observable<BookmarksPerPage[]>;
+  bookmarks: Bookmark[];
+  documentId: string;
+
+  private $subscription: Subscription;
```
benouaer, line 20: you're not setting this, so no need for it

projects/media-viewer/src/lib/bookmark/services/bookmark-api.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Observable } from 'rxjs';
+import { HttpClient } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+
+// @todo check if I can juts ref the BookmarkApiService in the annotations module instead?
+@Injectable()
+export class BookmarkApiService {
```
benouaer, line 7: there's already a bookmarks-api service that has this endpoint
https://github.com/hmcts/media-viewer/blob/master/projects/media-viewer/src/lib/annotations/bookmarks-api.service.ts

projects/media-viewer/src/lib/store/reducers/bookmarks.reducer.ts
```diff
@@ -19,47 +20,61 @@ export function bookmarksReducer (state = initialBookmarksState,
...
     }
 
     case fromBookmarks.CREATE_BOOKMARK_SUCCESS: {
       const bookmark: Bookmark = action.payload;
       const bookmarkEntities = {
         ...state.bookmarkEntities,
         [bookmark.id]: bookmark,
-      }
+      };
+      const bookmarkArray = Object.keys(bookmarkEntities).map(key => bookmarkEntities[key]);
```
benouaer, line 46: we should already have the bookmarks array (bookmarks), that's what the backend returns

projects/media-viewer/src/lib/store/reducers/bookmarks.reducer.ts
```diff
@@ -68,19 +83,23 @@ export function bookmarksReducer (state = initialBookmarksState,
         ...state.bookmarkEntities,
         [bookmark.id]: { ...bookmark }
       };
+      const bookmarkArray = Object.keys(bookmarkEntities).map(key => bookmarkEntities[key]);
```
benouaer, line 86: as above, we should already have the bookmark array

projects/media-viewer/src/lib/store/reducers/bookmarks.reducer.ts
```diff
@@ -19,47 +20,61 @@ export function bookmarksReducer (state = initialBookmarksState,
...
-      }
+      };
     }
 
     case fromBookmarks.DELETE_BOOKMARK_SUCCESS: {
       const bookmarkId: string = action.payload;
       const bookmarkEntities = { ...state.bookmarkEntities };
+      const page = bookmarkEntities[bookmarkId].pageNumber;
       delete bookmarkEntities[bookmarkId];
+      const pageBookmarkRemoved = [
```
benouaer, line 64: why not do the same as update below, and just regenerate the pageEntities

projects/media-viewer/src/lib/store/selectors/bookmarks.selectors.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      });
+      const payload: any = {
+        id: '1',
+        documentId: 'uuid',
+        name: 'bookmark1',
+        pageNumber: 0,
+        xCoordinate: 100,
+        yCoordinate: -120
+      };
+      store.dispatch(new fromActions.CreateBookmark(payload));
```
benouaer, line 44: do we need the CreateBookmark action dispatch here, since the test is for the selector? I think the action dispatch needed is the AddPages one, which we're already doing above

projects/media-viewer/src/lib/store/selectors/bookmarks.selectors.ts
```diff
@@ -30,12 +36,33 @@ export const getEditableBookmark = createSelector(
...
   fromAnnotations.getDocumentIdSetId,
   fromDocument.getPdfPosition,
-  (docSetId, pdfPosition) => {
+  fromDocument.getPages,
+  (docSetId, pdfPosition, pages) => {
     return {
       pageNumber: pdfPosition.pageNumber - 1,
       xCoordinate: pdfPosition.left,
-      yCoordinate: pdfPosition.top,
+      yCoordinate: pages[pdfPosition.pageNumber].styles.height - (pdfPosition.top * pages[pdfPosition.pageNumber].viewportScale),
```
benouaer, line 44: don't we need to do the same thing for the xCoordinate? does it not become relevant with rotation?

projects/media-viewer/src/lib/store/selectors/bookmarks.selectors.ts
```diff
@@ -30,12 +36,33 @@ export const getEditableBookmark = createSelector(
 export const getBookmarkInfo = createSelector(
   fromAnnotations.getDocumentIdSetId,
```
benouaer, line 37: I think it's better to use the document selector to get the documentId rather than annotations

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.html
```diff
@@ -46,6 +46,9 @@
                     [height]="pdfViewer.offsetHeight"
                     [pageHeights]="pageHeights">
     </mv-comment-set>
+    <mv-bookmark [zoom]="zoom"
```
benouaer, line 49: maybe rename this to reflect that it's the bookmark icon/placeholder, rather than the actual bookmark

projects/media-viewer/src/lib/store/reducers/bookmarks.reducer.ts
jasonpaige: I'm not sure what you mean? I think the only reason I'm loading them here is because they are needed in the new bookmarkPageEntities collection...

projects/media-viewer/src/lib/store/selectors/bookmarks.selectors.ts
jasonpaige: Not from what I've seen, the y coordinate is translated as part of the rotation. the x is still a fixed distance from the margin.

benouaer: ok, if it's working when the document is rotated, then that's fine

projects/media-viewer/src/lib/viewers/pdf-viewer/side-bar/bookmarks/bookmarks.component.ts
```diff
@@ -53,20 +53,32 @@ export class BookmarksComponent implements OnInit, OnDestroy {
   }
 
   goToBookmark(bookmark: Bookmark) {
+
+    const scaledY: (yCoordinate: number, height: number, page: DocumentPages, zoom: number) => number =
+      function(yCoordinate, height, page, zoom) {
```
benouaer, line 58: when does this function get called

jasonpaige: just below on line 108 and 116

PR #295, opened 2020-06-25
Refactor-ICP-Directive

meganaslattery:
### Change description ###

Refactor ICP code - convert directive to a service. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/media-viewer/src/lib/icp/icp.service.ts
+++ b/projects/media-viewer/src/lib/icp/icp.service.ts
@@ -27,11 +27,11 @@ export class IcpService implements OnDestroy  {
               private readonly socketService: IcpUpdateService,
               private readonly presenterSubscriptions: IcpPresenterService,
               private readonly followerSubscriptions: IcpFollowerService,
-              private store: Store<IcpState>) {}
-
-  setUp(id: string) {
-    this.caseId = id;
-    this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
+              private store: Store<IcpState>) {
+    this.store.pipe(select(fromIcpSelectors.getCaseId), filter(value => !!value)).subscribe(caseId => {
+      this.caseId = caseId;
+      this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
+    });
   }
 
   ngOnDestroy() {
--- a/projects/media-viewer/src/lib/icp/icp.service.ts
+++ b/projects/media-viewer/src/lib/icp/icp.service.ts
@@ -28,10 +28,12 @@ export class IcpService implements OnDestroy  {
               private readonly presenterSubscriptions: IcpPresenterService,
               private readonly followerSubscriptions: IcpFollowerService,
               private store: Store<IcpState>) {
-    this.store.pipe(select(fromIcpSelectors.getCaseId), filter(value => !!value)).subscribe(caseId => {
+    this.subscription = this.store.pipe(select(fromIcpSelectors.getCaseId), filter(value => !!value)).subscribe(caseId => {
       this.caseId = caseId;
-      this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
     });
+    this.subscription.add(this.toolbarEvents.icp.sessionLaunch.subscribe(() => {
+      if (this.caseId) { this.launchSession(); }
+    }));
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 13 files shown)


projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -121,6 +123,8 @@ export class PdfViewerComponent implements AfterContentInit, OnChanges, OnDestro
       this.pdfWrapper.positionUpdated.asObservable().pipe(throttleTime(500, asyncScheduler, { leading: true, trailing: true }))
       .subscribe(event => this.store.dispatch(new PdfPositionUpdate(event.location)))
     );
+
+    if (this.enableICP) { this.icpService.setUp(this.caseId); }
```
benouaer, line 127: I think it would make sense to get the caseId from the store

projects/media-viewer/src/lib/icp/icp.service.ts
```diff
@@ -35,7 +29,8 @@ export class IcpDirective implements OnInit, OnDestroy {
               private readonly followerSubscriptions: IcpFollowerService,
               private store: Store<IcpState>) {}
 
-  ngOnInit() {
+  setUp(id: string) {
```
benouaer, line 32: should probably get the caseId from the store

projects/media-viewer/src/lib/icp/icp.service.ts
```diff
@@ -35,7 +29,8 @@ export class IcpDirective implements OnInit, OnDestroy {
...
+    this.caseId = id;
```
benouaer, line 33: should probably get the caseId from the store

fazanki:
thank you :)

projects/media-viewer/src/lib/icp/icp.service.ts
```diff
@@ -33,10 +27,11 @@ export class IcpDirective implements OnInit, OnDestroy {
               private readonly socketService: IcpUpdateService,
               private readonly presenterSubscriptions: IcpPresenterService,
               private readonly followerSubscriptions: IcpFollowerService,
-              private store: Store<IcpState>) {}
-
-  ngOnInit() {
-    this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
+              private store: Store<IcpState>) {
+    this.store.pipe(select(fromIcpSelectors.getCaseId), filter(value => !!value)).subscribe(caseId => {
```
benouaer, line 31: this should probably also be added to the subscription, so that it can also be destroyed

projects/media-viewer/src/lib/icp/icp.service.ts
```diff
@@ -33,10 +27,11 @@ export class IcpDirective implements OnInit, OnDestroy {
...
+      this.caseId = caseId;
+      this.subscription = this.toolbarEvents.icp.sessionLaunch.subscribe(() => this.launchSession());
```
benouaer, line 33: not sure we need to nest the two subscriptions, we can just check if caseId is set

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -12,6 +12,15 @@ export function icpReducer (state = initialIcpSessionState,
 
   switch (action.type) {
 
+    case fromIcpActions.SET_CASE_ID: {
+      const caseId = action.payload;
+      const session = {...state.session, caseId};
```
benouaer, line 17: wouldn't the caseId exist before the session is started?

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -121,6 +126,7 @@ export class PdfViewerComponent implements AfterContentInit, OnChanges, OnDestro
       this.pdfWrapper.positionUpdated.asObservable().pipe(throttleTime(500, asyncScheduler, { leading: true, trailing: true }))
       .subscribe(event => this.store.dispatch(new PdfPositionUpdate(event.location)))
     );
+    if (this.caseId) { this.icpStore.dispatch(new SetCaseId(this.caseId)); }
```
benouaer, line 129: I would put this in ngOnChanges

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
meganaslattery: Yes.. I'm storing the caseId in the session part of the state tree because the session was already:
session: { caseId: string, sessionId: string, date: Date }

PR #304, opened 2020-07-23
Em 2695 icp participant list

meganaslattery:

https://tools.hmcts.net/jira/browse/EM-2695

### Change description ###

View participants list

```
[ ] Yes
[x] No
```


projects/media-viewer/src/lib/annotations/comment-set/comment-set.component.ts
```diff
@@ -63,6 +63,11 @@ export class CommentSetComponent implements OnInit, OnDestroy, OnChanges {
         this.showCommentsPanel = toggle;
       })
     );
+    this.subscriptions.push(this.toolbarEvents.icp.icpParticipantsListVisible.subscribe(isVisible => {
```
benouaer, line 66: can use filter here, so we only subscribe if it's true

projects/media-viewer/src/lib/icp/participants-list/participants-list.component.html
```diff
@@ -0,0 +1,18 @@
...
+  <div class="participant__card no-shadow">
+    <p class="participant__name">All Participants</p>
+  </div>
+
+  <div *ngIf="(presenter$ | async)?.username !== ''" class="participant__card">
+    <div class="participant__header"> Presenting </div>
+    <p class="participant__name">{{(presenter$ | async)?.username}}</p>
+  </div>
+
+  <div *ngIf="!((participants$ | async).length == 1 && (isPresenter$ | async))" class="participant__card">
```
benouaer, line 12: change into a positive statement

projects/media-viewer/src/lib/icp/participants-list/participants-list.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(private readonly toolbarEvents: ToolbarEventService,
+              private store: Store<IcpState>) {}
+
+  ngOnInit() {
+    this.participants$ = this.store.pipe(select(fromSelectors.getParticipants));
+    this.presenter$ = this.store.pipe(select(fromSelectors.getPresenter));
+    this.isPresenter$ = this.store.pipe(select(fromSelectors.isPresenter));
+
+    this.subscription = this.toolbarEvents.icp.icpParticipantsListVisible.subscribe(isVisible => this.showParticipantsList = isVisible);
+    this.subscription.add(this.toolbarEvents.commentsPanelVisible.subscribe(isVisible => {
```
benouaer, line 30: I think the mutual exclusivity of comments-panel and participants list should be managed at the toolbar-events service level, and rather than the components

projects/media-viewer/src/lib/store/reducers/icp.reducer.ts
```diff
@@ -28,10 +29,22 @@ export function icpReducer (state = initialIcpSessionState,
...
         client: participantInfo.client,
-        presenter: participantInfo.presenter
+        presenter: participantInfo.presenter,
       };
     }
 
+    case fromIcpActions.ICP_PARTICIPANT_LIST_UPDATED: {
+      const updatedParticipants: any = action.payload;
+      const participants = [];
+      for (const id in updatedParticipants) {
```
benouaer, line 39: this could be done in a more functional way
const participants = Object.keys({...updatedParticipants}).map(id => ({ id: id, username: updatedParticipants[id] }))

projects/media-viewer/src/lib/toolbar/icp-event.service.ts
```diff
@@ -6,6 +6,7 @@ import { BehaviorSubject, Subject } from 'rxjs';
   export const stoppingPresenting = new Subject();
   export const leavingSession = new BehaviorSubject(false);
   export const sessionExitConfirmed = new Subject();
+  export const icpParticipantsListVisible = new BehaviorSubject(false);
```
benouaer, line 9: participantsListVisible should be enough, I think

projects/media-viewer/src/lib/toolbar/icp-event.service.ts
```diff
@@ -30,5 +31,10 @@ import { BehaviorSubject, Subject } from 'rxjs';
 
   export const confirmExit = () => {
     sessionExitConfirmed.next();
+    icpParticipantsListVisible.next(false);
     enabled.next(false);
   };
+
+  export const toggleIcpParticipantsList = (isVisible) => {
```
benouaer, line 38: I think this is where we should set the comments-panel to false

projects/media-viewer/src/lib/toolbar/icp-toolbar/icp-toolbar.component.html
```diff
@@ -35,4 +35,14 @@
       </button>
     </div>
   </div>
+  <div id="toolbarViewerRight">
+    <div class="splitToolbarButton">
+      <button class="toolbarButton"
+              id="icp-participants-list" tabindex="0" title="Participants list"
+              (click)="showParticipantsList()">
+        <span *ngIf="!(toolbarEventService.icp.icpParticipantsListVisible | async)" data-l10n-id="icp-stop-presenting_label">Show Participants</span>
```
benouaer, line 43: only thing that's different here is the text, so we should probably use the if statement on that
{{ 'Hide Participants' ? (toolbarEventService.icp.icpParticipantsListVisible | async) : 'Show participants'  }}

PR #316, opened 2020-11-02
Media Viewer GDS consistency amends

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-2371
https://tools.hmcts.net/jira/browse/EUI-2392

### Change description ###
Update media viewer toolbar to make GDS-consistent (including 'More options'). Includes button visibility when opening documents that are not supported/have an error.

```
[ ] Yes
[x] No
```


projects/media-viewer/src/lib/shared/directives/integer-input-directive.ts
```diff
@@ -0,0 +1,23 @@
+import { Directive, ElementRef, HostListener, Self } from '@angular/core';
```
benouaer, line 1: is this actually needed? would <input type="number"> not do the same job?

paulhowes-HMCTS: I wanted to prevent the user from using the up/down arrows when inside the page number field so thought I'd add this, however on reflection I will remove this directive and simply add css to remove the up/down arrows that appear inside a number input field as there are already page up and down buttons on the toolbar.

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
```diff
@@ -1,22 +1,14 @@
...
-      <mv-tb-left-pane></mv-tb-left-pane>
-      <mv-tb-middle-pane></mv-tb-middle-pane>
-      <mv-tb-right-pane
-        [enableRedaction]="enableRedactions"
-        [enableICP]="enableICP"
-        [enableAnnotations]="enableAnnotations"
-      >
-      </mv-tb-right-pane>
-    </div>
+    <mv-media-viewer-toolbar [enableRedaction]="enableRedactions" [enableICP]="enableICP"
```
benouaer, line 4: I think we can combine the main-toolbar and the media-viewer-toolbar, otherwise it's not clear what does what from the names

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
```diff
@@ -0,0 +1,193 @@
+<div class="mv-toolbar__container">
```
benouaer, line 1: it seems like you could split this into 3 components to simplify the html: mainToolbar, moreOptions, menuItems

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  @HostListener('window:resize', [])
+  public onResize() {
+    this.cdr.detectChanges();
+    // Call calculation method
+    this.checkCalculation();
+  }
+
+  private checkCalculation() {
```
benouaer, line 127: this is way too complex for what it's doing

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.ts
```diff
@@ -0,0 +1,281 @@
...
+          break;
+        }
+      }
+
+      // Call this method to recalculate for rest of toolbar elements
+      this.checkCalculation();
+    }
+  }
+
+  public isToolbarButtonVisible(buttonId: string): boolean {
```
benouaer, line 168: if we're gonna use this, it should be a map rather than an array, so that we can go directly to the button we want

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.ts
```diff
@@ -0,0 +1,281 @@
...
+    for (const buttonState of this.toolbarButtonVisibilityStates) {
+      if (buttonState.id === buttonId) {
+        return buttonState.showOnToolbar;
+      }
+    }
+  }
+
+  private setToolbarButtonVisible(buttonId: string, visible: boolean) {
```
benouaer, line 176: same as comment above, should be map not an array

projects/media-viewer/src/lib/toolbar/search-bar/search-bar.component.ts
```diff
@@ -81,15 +83,22 @@ export class SearchBarComponent implements OnInit, OnDestroy {
   }
 
   private setSearchResultsCount(results: SearchResultsCount): void {
-    this.haveResults = results.total > 0;
-    this.resultsText = this.haveResults
-      ? `${results.current} of ${results.total} matches`
-      : 'Phrase not found';
+    this.resultCount = results.total;
```
benouaer, line 86: is the resultCount variable needed, given we can just use results.total?

projects/media-viewer/src/lib/toolbar/search-bar/search-bar.component.html
```diff
@@ -1,39 +1,60 @@
...
-      </button>
-      <div class="splitToolbarButtonSeparator"></div>
-      <button id="findNext" class="toolbarButton findNext" title="Find the next occurrence of the phrase"
-              tabindex="0" data-l10n-id="find_next" (click)="searchNext()">
-        <span data-l10n-id="find_next_label">Next</span>
+<div class="searchbar govuk-!-padding-3" [hidden]="toolbarEvents.searchBarHidden | async">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <input class="govuk-input govuk-!-display-inline-block govuk-!-width-three-quarters govuk-!-margin-bottom-1"
+        type="text" #findInput [(ngModel)]="searchText" (keydown)="onInputKeyPress($event)" title="Find"
```
benouaer, line 5: can we filter on the keys we want here instead of listening for all keys, so we would just do (keydown.space), (keydown.escape), (keydown.enter), with a corresponding method for each

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.html
```diff
@@ -1,64 +1,49 @@
-<div id="secondaryToolbar" class="secondaryToolbar doorHangerRight"
-     [hidden]="toolbarEvents.subToolbarHidden | async">
+<div id="secondaryToolbar" class="secondaryToolbar doorHangerRight" [hidden]="toolbarEvents.subToolbarHidden | async">
```
benouaer, line 1: is the subtoolbar still needed given you're using the overlay? maybe you could reuse it for the overlay stuff

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.ts
```diff
@@ -49,7 +49,7 @@ export class SubToolbarComponent implements OnInit, OnDestroy {
     this.closeMenu();
   }
 
-  enterIcpMode() {
+  togglePresentBar() {
```
benouaer, line 52: if we rename this we should rename it everywhere to be consistent

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
```diff
@@ -30,7 +31,8 @@ export class ToolbarButtonVisibilityService {
     this.showNavigation = false;
     this.showZoom = false;
     this.showRotate = false;
-    this.showPresentationMode = false;
+    this.showPresentationMode = true;
```
benouaer, line 34: options should not be set to true here, because this method is resetting all the options, so that they can be set with the options object being passed in

projects/media-viewer/src/lib/toolbar/toolbar.module.ts
```diff
@@ -3,29 +3,25 @@ import { CommonModule } from '@angular/common';
...
-import { ToolbarRightPaneComponent } from './right-pane/right-pane.component';
-import { ToolbarMiddlePaneComponent } from './middle-pane/middle-pane.component';
 import { FormsModule } from '@angular/forms';
 import { ToolbarButtonVisibilityService } from './toolbar-button-visibility.service';
 import { ToolbarEventService } from './toolbar-event.service';
 import { RedactionToolbarComponent } from './redaction-toolbar/redaction-toolbar.component';
- import { IcpToolbarComponent } from './icp-toolbar/icp-toolbar.component';
-
+import { IcpToolbarComponent } from './icp-toolbar/icp-toolbar.component';
+import { MediaViewerToolbarComponent } from './media-viewer-toolbar/media-viewer-toolbar.component';
```
benouaer, line 11: as mentioned in another comment, this should probably be combined with the main-toolbar component, which can be split into two, one for the horizontal-view and one for the vertical-viewe

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
```diff
@@ -0,0 +1,193 @@
...
+            <button [disabled]="toolbarEvents.zoomValueSubject.value == 0.1" id="mvMinusBtn"
+              class="mv-toolbar__menu-button--zoom-out button-image" title="Zoom Out" data-l10n-id="zoom_out"
+              (click)="stepZoom(-0.1)"><span></span></button>
+            <select id="scaleSelect" class="hmcts-toolbar-select" title="Zoom" tabindex="0" data-l10n-id="zoom"
+              (change)="zoom($event.target.value)" aria-label="zoom">
+              <option #zoomSelect id="customScaleOption" title="" [value]="toolbarEvents.zoomValueSubject.value">
+                {{(toolbarEvents.zoomValueSubject.value) * 100 | number: '1.0-0'}}%</option>
+              <option title="" value="0.1" data-l10n-id="page_scale_percent" data-l10n-args='{ "scale": 10 }'>10%
+              </option>
+              <option title="" value="0.25" data-l10n-id="page_scale_percent" data-l10n-args='{ "scale": 25 }'>25%
```
benouaer, line 86: should this be a for loop?

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
```diff
@@ -0,0 +1,193 @@
...
+          id="mvCommentsBtn" #mvCommentsBtn class="mv-button mv-toolbar__menu-button--comments" title="Comments"
+          data-l10n-id="comments" [class.button-hidden-on-toolbar]="!isToolbarButtonVisible(mvCommentsBtn.id)"
+          [class.button-hidden-on-dropdown]="isToolbarButtonVisible(mvCommentsBtn.id)" aria-pressed="false"
+          (click)="toggleCommentsPanel()">
+          <span>Comments</span>
+        </button>
+
+        <!-- TODO: Where is this and how does it work?
+      <ng-container *ngIf="!icpEnabled">
+        <a *ngIf="toolbarButtons.showBookmark"
```
benouaer, line 182: what is this?

projects/media-viewer/src/lib/toolbar/search-bar/search-bar.component.ts
paulhowes-HMCTS: resultCount is used by the HTML template to hide the next / prev links if there are no search results found

projects/media-viewer/src/lib/toolbar/search-bar/search-bar.component.html
paulhowes-HMCTS: I've updated the code and this will be available in the next commit.

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.ts
paulhowes-HMCTS: I have?

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
paulhowes-HMCTS: That's how it was originally, however I have put it in an ngFor loop and set the values from the ts file.

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
paulhowes-HMCTS: As discussed, this is legacy code and will be removed in the next commit.

projects/media-viewer/src/lib/toolbar/main-toolbar/main-toolbar.component.html
paulhowes-HMCTS: Agreed. This will be resolved in the next commit.

projects/media-viewer/src/lib/toolbar/sub-toolbar/sub-toolbar.component.html
paulhowes-HMCTS: The subtoolbar will be removed as part of the next commit HOWEVER the tests relating to the zoom, print, download and rotate will be moved to the maintoolbar as these are still required. Note the secondary button tests are no longer required as there is only one toolbar driven by the template.

projects/media-viewer/src/lib/toolbar/toolbar.module.ts
paulhowes-HMCTS: Agreed. This will be combined with the main-toolbar in the next commit.

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.html
paulhowes-HMCTS: Related to code complexity. Potential for additional technical debt if a different approach is adopted.

projects/media-viewer/src/lib/toolbar/media-viewer-toolbar/media-viewer-toolbar.component.ts
paulhowes-HMCTS: Related to code complexity. Potential for additional technical debt if a different approach is adopted.

benouaer (approved):

benouaer (dismissed):

projects/media-viewer/src/lib/toolbar/toolbar-button-visibility.service.ts
```diff
@@ -72,13 +74,19 @@ export const defaultImageOptions = {
...
 };
 
 /**
  * Default toolbar state for unsupported media
  */
 export const defaultUnsupportedOptions = {
   showDownload: true,
-  showCommentSummary: false
+  showCommentSummary: false,
+  showDrawButton: false,
```
benouaer, line 87: you shouldn't need to add any false options here, because everything is false by default, and you just set things to true if you want them to appear

benouaer (approved):

PR #324, opened 2021-01-04
fix counter clockwise image rotation

benouaer:
https://tools.hmcts.net/jira/browse/EM-3256

### Change description ###
- fix counter clockwise image rotation by using 270 instead of -90

```
[ ] Yes
[X] No
```


paulhowes-HMCTS (approved):
-ve rotation appeared to work when tested earlier, however +270 rotation has the same effect.

rajkatla-hmcts (approved):

PR #337, opened 2021-02-08
Bug/em 2902 no horizontal scrolling

yyaremchuk:

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)

### Change description ###

```
[ ] Yes
[ ] No
```


projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -102,6 +125,8 @@ export class MediaViewerComponent implements OnChanges, OnDestroy, AfterContentI
       .subscribe(changes => this.onCommentChange(changes));
     this.$subscriptions.add(this.toolbarEvents.getShowCommentSummary()
       .subscribe(changes => this.showCommentSummary = changes));
+
+    //
```
benouaer, line 129: I guess this is not needed

projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -50,13 +52,33 @@ enum ConvertibleContentTypes {
...
   @Input() showToolbar = true;
   @Input() toolbarButtonOverrides: any = {};
-  @Input() height = 'calc(100vh - 42px)';
+
+  @Input()
+  public set height(value: string) {
+    this._givenHeight = value;
+  }
+
+  public get height(): string {
```
benouaer, line 69: I think it would be better if we don't use the getter function in this way, because we're using the field in the template and it's going to be called every time the angular change detection runs. when in reality, the value only changes on zoom, rotate or when the input height changes
so using a function that only gets called at those times would be better in terms of performance

yyaremchuk: That's all true. I just wanted to avoid introducing a new property, for example internalHeight or something like this, which can be consumed by the template and updated when height input param changes.

benouaer: but that would probably be better in terms of performance
there are quite a few instances where we have done this in the app, and we're trying to get rid of them

yyaremchuk: Removed setter/getter for height input param and implement AfterViewChecked hook to track component location change. Tried to male it lean and not to trigger change detection on every run.

projects/media-viewer/src/lib/media-viewer.component.ts
```diff
@@ -80,19 +90,25 @@ export class MediaViewerComponent implements OnChanges, OnDestroy, AfterContentI
...
+    private elRef: ElementRef,
+    private cdr: ChangeDetectorRef
   ) {
     if (this.annotationApiUrl) {
       api.annotationApiUrl = this.annotationApiUrl;
     }
   }
 
+  public ngOnInit(): void {
+  }
```
benouaer, line 110: I assume this is not needed

benouaer (dismissed):

benouaer (dismissed):

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.html
```diff
@@ -15,42 +15,59 @@
...
-  ></mv-redactions>
-
-  <ng-template #annotationTemplate>
-    <mv-metadata-layer
-      *ngIf="enableAnnotations && annotationSet"
-      [zoom]="zoom"
-      [rotate]="rotation">
-    </mv-metadata-layer>
+  <div #scrollTwo class="comments">
     <mv-comment-set #commentsPanel
```
benouaer, line 60: would something like this not work? [scrollTop]="viewerContainer.scrollTop"

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -121,7 +118,16 @@ export class PdfViewerComponent implements AfterContentInit, OnChanges, OnDestro
       this.toolbarEvents.changePageByDeltaSubject.subscribe(pageNumber => this.pdfWrapper.changePageNumber(pageNumber))
     );
     this.$subscription.add(this.toolbarEvents.grabNDrag.subscribe(grabNDrag => this.enableGrabNDrag = grabNDrag));
-    this.$subscription.add(this.toolbarEvents.commentsPanelVisible.subscribe(toggle => this.showCommentsPanel = toggle));
+    this.$subscription.add(this.toolbarEvents.commentsPanelVisible.subscribe(toggle => {
+        this.showCommentsPanel = toggle;
+
+        if (toggle) {
```
benouaer, line 124: would setting the property in the html like below not work?
[scrollTop]="viewerContainer.scrollTop"

projects/media-viewer/src/lib/viewers/pdf-viewer/pdf-viewer.component.html
yyaremchuk: Potentially, yes but we want to scroll commentsPanel when it's visible and the viewer is scrolled. not on every digest cycle

benouaer (approved):

PR #381, opened 2021-07-28
EM-3566-Download E2E

phani-perla1:
https://tools.hmcts.net/jira/browse/EM-3566

### Change description ###
Download E2E

```
[ ] Yes
[ x] No
```

### DIFF
--- a/e2e/src/pages/app.po.ts
+++ b/e2e/src/pages/app.po.ts
@@ -67,6 +67,15 @@ export class AppPage {
     }, 10000, 'failed to load search results');
   }
 
+  async clickAction(key: string) {
+    const downLoadRef = element.all(by.id('mvDownloadBtn')).filter(function (elm) {
+      return elm.isDisplayed().then(function (isDisplayed) {
+        return isDisplayed;
+      });
+    }).last();
+    downLoadRef.click();
+  }
+
   async waitForElementsArray(selector: Locator) {
     await browser.wait(async () => {
       return (await element.all(selector).isPresent());
--- a/e2e/src/pages/app.po.ts
+++ b/e2e/src/pages/app.po.ts
@@ -67,15 +67,6 @@ export class AppPage {
     }, 10000, 'failed to load search results');
   }
 
-  async clickAction(key: string) {
-    const downLoadRef = element.all(by.id('mvDownloadBtn')).filter(function (elm) {
-      return elm.isDisplayed().then(function (isDisplayed) {
-        return isDisplayed;
-      });
-    }).last();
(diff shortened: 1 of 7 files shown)


e2e/src/pages/download.po.ts
```diff
@@ -15,4 +16,12 @@ export class DownloadPage extends AppPage {
   hasFileDownloaded(filePath: string) {
     return fs.existsSync(filePath);
   }
+
+  async clickMoreOptions() {
+    await this.clickElement(by.id('mvMoreOptionsBtn'));
+  }
+
+  async clickDownloadButton() {
```
benouaer, line 24: use xpath like can be problematic, because any changes in the document structure would break the test unnecessarily
I would suggest using a class or id to select the element

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -113,6 +119,13 @@ Then('I expect bookmark to be added to the existing list', async function () {
   this.attach(screenshots, 'image/png');
 });
 
+Then('I expect to see the document should be downloaded', async function () {
```
benouaer, line 122: we're not actually checking whether the document has been downloaded

benouaer (changes requested):
select element by id or class instead of xpath

e2e/src/step_definitions/media-viewer.ts
phani-perla1: When the script clicks download option,  the document is downloading and at that stage we are capturing screenshoot.

e2e/src/pages/download.po.ts
phani-perla1: I agree with you. Added generic method that will take care when the object has same reference type.
Same reference has been given  for download, that's causing the issue. id="mvDownloadBtn"

e2e/src/pages/app.po.ts
```diff
@@ -67,6 +67,15 @@ export class AppPage {
     }, 10000, 'failed to load search results');
   }
 
+  async clickAction(key: string) {
+    const downLoadRef = element.all(by.id('mvDownloadBtn')).filter(function (elm) {
```
benouaer, line 71: not sure what this is? are you just trying to click the download button? does this not work
this.clickElement(by.id('mvDownloadBtn'))

phani-perla1 (commented):
Addressed reviewer comments

phani-perla1:
> select element by id or class instead of xpath

Xpath issue fixed

benouaer (approved):

PR #384, opened 2021-08-06
EM-400-Multimedia E2E Tests

phani-perla1:

https://tools.hmcts.net/jira/browse/EM-4000

### Change description ###

1. Multimedia Video E2E Tests
2. 

```
[ ] Yes
[ x] No
```


e2e/src/features/multimedia.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Background:
+    Given I am on Media Viewer Page
+    Then I expect the page header to be "MEDIA VIEWER DEMO"
+    When the user selects the multimedia option
+
+  @EM-4000 @play
+  Scenario: Check whether the video player is working or not?
+    And the user selects play option
+    Then I should see video should be in play mode
```
benouaer, line 12: this should probably be "I should see video in play mode"

e2e/src/features/multimedia.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  @EM-4000 @pause
+  Scenario: Pause Video
+    And the user selects pause option
+    Then I should see video should be in pause mode
```
benouaer, line 17: same thing here "I should see video in pause mode"

e2e/src/pages/multimedia.po.ts
```diff
@@ -0,0 +1,17 @@
...
+import {by} from 'protractor';
+
+export class MultimediaPage extends AppPage {
+
+  async clickMultimedia() {
+    await this.clickElement(by.id('multimedia-button'));
+  }
+
+  async clickPlayButton() {
+    await this.clickElement(by.css('#mainContainer > div > mv-multimedia-player'));
```
benouaer, line 11: pause and play are exactly the same, so why create two functions?
also, we can just use this selector, can't we "mv-multimedia-player"?

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -127,6 +127,32 @@ Then('I expect to see the document should be downloaded', async function () {
   await downloadPage.waitForDownloadToComplete(file);
 });
```
benouaer, line 129: can we put the multimedia functions in their own file (we can separate the other functionality as we go along)
this file is getting too big, and it shouldn't have everything in it. each feature should have its own file to make it more manageable

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -127,6 +127,32 @@ Then('I expect to see the document should be downloaded', async function () {
...
 
+When('the user selects the multimedia option', async () => {
+  await multimediaPage.clickMultimedia();
+  await genericMethods.sleep(5000);
```
benouaer, line 132: do we need the sleep method? adding 5secs to every click would really slow down the tests
I thought the clickElement method already has waiting time in it

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -127,6 +127,32 @@ Then('I expect to see the document should be downloaded', async function () {
...
+});
+
+When('the user selects play option', async () => {
+  await multimediaPage.clickPlayButton();
+});
+
+Then('I should see video should be in play mode', async function ()  {
```
benouaer, line 139: if we're just taking a screenshot this becomes a manual test rather than automated
is there anything in the markup we can check to see if the video is playing?
from what I could see, the play button isn't clickable if the video can't be played

e2e/src/step_definitions/media-viewer.ts
```diff
@@ -127,6 +127,32 @@ Then('I expect to see the document should be downloaded', async function () {
...
+    await genericMethods.sleep(5000);
+    const screenshots = await browser.takeScreenshot();
+    this.attach(screenshots, 'image/png');
+});
+
+When('the user selects pause option', async () => {
+  await multimediaPage.clickPlayButton();
+});
+
+Then('I should see video should be in pause mode', async function () {
```
benouaer, line 149: again here, we're not really checking here, which means manual intervention is required to validate if the test has passed

benouaer (changes requested):
some changes are needed, see comments

e2e/src/features/multimedia.feature
phani-perla1: fixed

e2e/src/features/multimedia.feature
phani-perla1: fixed

e2e/src/pages/multimedia.po.ts
phani-perla1: fixed.
Yes we can use css selector ref as "mv-multimedia-player"

e2e/src/step_definitions/media-viewer.ts
phani-perla1: sure

e2e/src/step_definitions/media-viewer.ts
phani-perla1: Let me check...

e2e/src/step_definitions/media-viewer.ts
phani-perla1: if ( pause) click success then video is in stop mode  and capturing the screenshot and that is attached to report .
else it throws error msg

e2e/src/step_definitions/media-viewer.ts
phani-perla1: Don't find the solution. That's the optimistic solution at the moment and we can improve it later

yogesh-hullatti (approved):

benouaer (approved):

PR #699, opened 2023-03-27
Em 4599 fix overlapping comments section

Jefex7:

[https://tools.hmcts.net/jira/browse/EM-4599](url)

### Change description ###

Added a margin to comment to fix overlapping comments section

```
[ ] Yes
[x] No
```


projects/media-viewer/src/lib/annotations/annotations.module.ts
Jefex7, line 1: I had an Issue where the provider was missing, I am not sure if this is the correct way to do it.

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-set-header.component.ts
```diff
@@ -1,29 +1,32 @@
-import { Component, EventEmitter, Input, OnDestroy, OnInit, Output, ViewEncapsulation } from '@angular/core';
+import { Component, EventEmitter, Inject, Input, OnDestroy, OnInit, Output, ViewEncapsulation } from '@angular/core';
```
MunishSharmaHMCTS, line 1: why do we have inject here

projects/media-viewer/src/lib/annotations/comment-set/comment/comment.service.ts
MunishSharmaHMCTS, line 1: might need unit test for these. Sonar cube won't pass

projects/media-viewer/src/lib/annotations/comment-set/comment-set-header/comment-set-header.component.ts
Jefex7: That is a mistake I forgot to remove it.

MunishSharmaHMCTS (approved):

yogesh-hullatti (approved):

PR #1203, opened 2025-09-02
EXUI-3453 - Update redaction page draw code to only run if redaction enabled

Josh-HMCTS:
### Jira link

See [EXUI-3453](https://tools.hmcts.net/jira/browse/EXUI-3453)

### Change description

Update the code that updates the page css to only run if redactions is enabled. this should resolve the performance issues affecting some users

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/media-viewer",
-  "version": "4.1.4-performance-issue",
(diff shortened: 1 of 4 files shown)


olusegz07 (approved):

connorpgpmcelroy (approved):
LGTM

